PR #2, opened 2020-01-19
Enable verilator build to support FST tracing

mballance:
In order for Verilator to support FST tracing with Conda, the package needs to call out 'zlib' as a dependency, and the tracing file needs to define __STD_FORMAT_MACROS. This patch adds these changes, and ensures they are tested by the build tests.

Signed-off-by: Matthew Ballance <matt.ballance@gmail.com>


mithro:
Looks like this doesn't break verilator, so I'm going to merge.

PR #12, opened 2020-02-06
Fix Yosys for upstream changes

mithro:


mithro:
@PiotrZierhoffer -- This pull request is obsolete now, right? We can delete it?

PR #14, opened 2020-04-22
pnr/nextpnr/ecp5: fix linux target

kowalewskijan:
PR changes:
- packages configuration in _meta.yaml_
- minor formatting fixes

### DIFF
--- a/pnr/nextpnr/ecp5/meta.yaml
+++ b/pnr/nextpnr/ecp5/meta.yaml
@@ -23,33 +23,32 @@ requirements:
   build:
     - {{ compiler('c') }}
     - {{ compiler('cxx') }}
+    - fontconfig  [not win]
     - cmake
-    - fontconfig
-    - python
-    - boost {{ boost_version }}
+    - python    {{ python }}
+    - boost     {{ boost_version }}
     - boost-cpp {{ boost_version }}
-    - libboost {{ boost_version }}
-    - py-boost {{ boost_version }}
+    - libboost  {{ boost_version }}
+    - py-boost  {{ boost_version }}
   host:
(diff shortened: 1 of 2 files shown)


pnr/nextpnr/ecp5/meta.yaml
```diff
@@ -23,33 +23,32 @@ requirements:
...
-    - libgcc-ng
-    - libstdcxx-ng
+    - libgcc-ng     [linux]
+    - libstdcxx-ng  [linux]
     - prjtrellis
     - {{ pin_compatible('python', min_pin='x.x', max_pin='x.x') }}
-    - libboost {{ boost_version }}
-    - py-boost {{ boost_version }}
+    - libboost {{ boost_version }} [not win]
+    - py-boost {{ boost_version }} [not win]
```
mithro, line 51: Why is boost `[not win]`?

kowalewskijan: This is the same reason why we have `[not win]` in `ice40` target ([link](https://github.com/litex-hub/litex-conda-eda/blob/master/pnr/nextpnr/ice40/meta.yaml#L53-L54)). Boost software should be installed via `vckpg` like [here](https://github.com/litex-hub/litex-conda-eda/blob/master/pnr/nextpnr/ice40/bld.bat#L11). The original reason why we use `vckpkg` was that Conda does not provide boost compiled with mingw. So using `boost` from conda only resulted in compliation errors.

kowalewskijan:
I moved these changes to [Travis fixes PR](https://github.com/litex-hub/litex-conda-eda/pull/17)

PR #15, opened 2020-04-22
pnr/nextpnr/ecp5: add macOS target

kowalewskijan:
PR changes:
- add OSX support for _nextpnr/ecp5_ package
- ~~add _DOT_ON_STDOUT_ functionality to avoid 10min timout from Travis~~

~~Marked as WIP due to _nextpnr/ecp5_ package hits 50min timeout on Travis for OSX.~~

### DIFF
--- a/.github/workflows/build-packages.yml
+++ b/.github/workflows/build-packages.yml
@@ -328,6 +328,19 @@ jobs:
           submodules: 'true'
       - uses: hdl/conda-ci@master
 
+  #73
+  nextpnr-ecp5-osx:
+    runs-on: "macos-latest"
+    needs: ["prjtrellis-osx", "symbiflow-yosys-3_2_1-osx", "symbiflow-yosys-3_3-osx"]
+    env:
+      PACKAGE: "pnr/nextpnr/ecp5"
+      OS_NAME: "osx"
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          submodules: 'true'
+      - uses: hdl/conda-ci@master
+
   #24
   nextpnr-generic-linux:
     runs-on: "ubuntu-16.04"


kowalewskijan:
I marked this package as "allow to fail"

.travis/dot.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
```
mithro, line 1: Travis already has a function which does this -- called something like `travis_wait`...

mithro: Also `travis_jigger` it seems -- https://github.com/travis-ci/travis-build/blob/master/lib/travis/build/bash/travis_wait.bash

mithro: https://github.com/chanzuckerberg/go-travis-wait and https://pypi.org/project/travis-wait-improved/ might be other options too...

kowalewskijan: Thanks for suggestions. I avoided `travis_wait` because it does not pipe the output during building. I'll look into `jigger` and `improved` version of `travis_wait`

mithro: @kowalewskijan -- This should probably be something in the common CI scripts too.

kowalewskijan: I think we can override default `travis_wait` and `travis_jigger` like here: https://github.com/matomo-org/travis-scripts/blob/master/travis-helper.sh and we can keep these overridden functions in `common.sh`.

kowalewskijan: I added overridden functions

mithro:
Does this need me to review?

PiotrZierhoffer:
I think we'll get this green and merge.

@kowalewskijan please take care to put these functions to litex-hub/litex-conda-ci (because of #25)

kowalewskijan:
I removed a commit associated with `KEEP_ALIVE` functionality since it will come with `litex-conda-ci`. We can merge this PR after #25

PR #16, opened 2020-05-18
Linux osx vtr and reworked channels

kowalewskijan:
PR contains following changes:
* rework adding custom channels via condarc (as it is in [litex-conda-misc](https://github.com/litex-hub/litex-conda-misc))
* add Linux target for VtR package
* add OSX target for VtR package


PiotrZierhoffer:
@mithro can we merge this?

PR #17, opened 2020-05-20
Travis fixes

kowalewskijan:
This PR contains a series of fixes for currently failing packages.


kowalewskijan:
Currently failing jobs are:
- `pnr/nextpnr/ecp5` - this should be covered by (https://github.com/litex-hub/litex-conda-eda/pull/14)
- `pnr/nextpnr/ice40` - this job is most likely related to `yosys` package somehow. I had a similar problem with conflicting packages when I worked on bundled [metapackage](https://github.com/SymbiFlow/conda-packages/pull/83) and I figured it out that the package conflicts appeared only when `yosys` was appended to metapackage. It was solved by adding `conda-forge` channel, but here it does not work.

I additionally added to this PR handling channels via condarc (like in `litex-conda-misc` repo).
I think we can merge these fixes and I'll investigate `pnr/nextpnr/ice40` package separately.

@mithro what do you think?

mithro:
This looks okay, merge if Travis goes green.

PR #18, opened 2020-05-27
Add `yosys-plugins`

JakubJatczak:
This PR adds `yosys-plugins` builds for `Win`, `Osx` and `Linux`

### DIFF
--- a/syn/yosys-plugins/fasm_build.patch
+++ b/syn/yosys-plugins/fasm_build.patch
@@ -1,0 +1,69 @@
+diff --git a/fasm-plugin/Makefile b/fasm-plugin/Makefile
+index 8c1a158..19a35bd 100644
+--- a/fasm-plugin/Makefile
++++ b/fasm-plugin/Makefile
+@@ -1,16 +1,15 @@
+ CXX = $(shell yosys-config --cxx)
+-CXXFLAGS = $(shell yosys-config --cxxflags)
+-LDFLAGS = $(shell yosys-config --ldflags)
+-LDLIBS = $(shell yosys-config --ldlibs)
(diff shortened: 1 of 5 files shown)


syn/yosys-plugins/fasm_build.patch
```diff
@@ -0,0 +1,69 @@
+diff --git a/fasm-plugin/Makefile b/fasm-plugin/Makefile
```
mithro, line 1: Can you just send this patch upstream?

JakubJatczak: Unfortunately it is Windows specific - mingw in conda compiles to `.o` file instead of `.so`.
Other change that is there, is converting paths to Unix format (basically changing '\' to '/'). This one won't harm other os compilation, but without full windows support it does not make sense to add it

mithro: The upstream plugin should support windows.

siriKeeton: This patch would cause Linux/Osx build failure. I will prepare portable Makefile or separate Makefile for Windows - whatever turns out to be cleaner solution.

JakubJatczak: This patch will cause build failure on Linux/Osx. I will prepare portable Makefile or separate Makefile for Windows - whatever turns out to be cleaner solution.

mithro:
Does this need me to review?

ajelinski:
Such package is already built on master as `symbiflow-yosys-plugins` for Linux and macOS. The Windows package can't be built yet because the `symbiflow-yosys` (which is the one this package should use instead of `yosys`) isn't built for Windows.

PR #19, opened 2020-06-08
sim/verilator: add OSX target

kowalewskijan:


mithro:
If it builds okay, feel free to merge.

PR #20, opened 2020-06-09
xilinx/vivado: add xilinx-vivado metapackage

kowalewskijan:
Package checks:
- If Vivado is in PATH
- If Vivado version matches the metapackage version


xilinx/vivado/pre-link.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+LOCAL_VIVADO_VERSION=`vivado -version | sed 's|.*Vivado v\([0-9\.]*\).*|\1|' | sed '1!d'`
+
+if [[ -z $LOCAL_VIVADO_VERSION ]]; then
+	echo "Please add Vivado to PATH by sourcing settings64.sh, then install package."
```
mithro, line 6: We don't want people to source `settings64.sh` as it mucks up the environment.

kowalewskijan: I corrected message to `echo "Please add Vivado to PATH then install package."`

mithro: Can we just check for common vivado install locations?

kowalewskijan: Ok, I think we can look for vivado in `PATH` first and then if it is not present we can look in common installation path: `/opt/Xilinx/Vivado/<version>/bin/vivado`

mithro: SGTM!

kowalewskijan: Fixed

mithro:
LGTM. Merge if green.

PR #30, opened 2020-07-01
Remove unneeded patch from magic

PiotrZierhoffer:
Tim Edwards merged the fix upstream


PiotrZierhoffer:
I'll merge it as soon as magic passes

PiotrZierhoffer:
And so it did.

PR #33, opened 2020-08-26
Symbiflow vtr

PiotrZierhoffer:
This is an attempt to get rid of https://github.com/litex-hub/litex-conda-eda/pull/24, which is too big to be merged


PiotrZierhoffer:
The only sensible way to approach it is to divide, as @mithro suggested. This patchset does vtr, next will be yosys and yosys-plugins. The last step will be to add QL packages

PiotrZierhoffer:
Same files are in sync, using symlinks. I did not use variants as I did not find the proper syntax to alter deps.

I do think these files will desync, as different forks will use different commits

PiotrZierhoffer:
There will also be a separate PR with a commit change for symbiflow-vtr, but I need to discuss the correct one with acomodi et al

PiotrZierhoffer:
Rebased onto current master, with the new stages layout

mithro (approved):
LGTM!

ajelinski:
All of the functional changes this PR had have already been merged to master through other PRs, including the aforementioned QL packages.

PR #36, opened 2020-10-28
Add `symbiflow-yosys-plugins` for Linux and macOS

ajelinski:


mithro:
LGTM!

PiotrZierhoffer:
@ajelinski these are symbiflow packets, so they should be called `symbiflow-yosys-plugins`

mithro:
These are "SymbiFlow created / owned" plugins for yosys tool, so I think all of `yosys-plugins-symbiflow`, `symbiflow-yosys-plugins` and `yosys-symbiflow-plugins` make sense.

I think we want this package to be group with the yosys package? IE If you hit <tab> on `yosys` you get the `yosys`, `yosys-plugins-symbiflow`, `yosys-plugins-ghdl`?

Thus, if the yosys package is called `symbiflow-yosys` then this package should be called `symbiflow-yosys-plugins-symbiflow`, if the yosys package is called `yosys` then this package should be called `yosys-plugins-symbiflow`?

PR #37, opened 2020-10-29
Add prjxray-db & prjxray-tools package

tjurtsch:
This recipe is based on 'Symbiflow/conda-packages' one.

fpga-tool-perf depends on prjxray-db
symbiflow-examples depends on both prjxray-db and prjxray-tools


mithro:
FYI -- @acomodi / @kgugala

PR #41, opened 2020-11-04
Add nextpnr-xilinx package

tjurtsch:
fpga-tool-perf depends on this package.

Recipe is based on 'Symbiflow/conda-packages' one.


tjurtsch:
There one change when compared to 'Symbiflow/conda-packages' recipe. The version of boost-related packages is fixed and defined in conda_build_config.yaml. Without setting the same version for all boost-related packages in meta.yaml we've had problems with nextpnr-xilinx being built with different boost than was later pulled in runtime (e.g. in TEST step in conda build).

PR #42, opened 2020-11-04
Add `quicklogic-vtr` and `quicklogic-vtr-gui` recipes for Linux

ajelinski:
These are exactly the same as in the current master of https://github.com/QuickLogic-Corp/ql_conda_eda

Difference with SymbiFlow versions of vtr lies mostly in using `master` instead of `master+wip` branch of the source repository.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -171,6 +171,16 @@ jobs:
     dist: xenial
     env:
     - PACKAGE=pnr/symbiflow-vtr-gui
+  - stage: "No dependencies"
+    os: linux
+    dist: xenial
+    env:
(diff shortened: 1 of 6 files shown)


ajelinski:
It turned out QuickLogic's `vtr` is compatible with the mainline `vtr` repository. Adding these packages while there is a `vtr` package built based on mainline repository seems pointless.

mithro:
@ajelinski -- Can we create a "dummy" package which just installs the upstream vtr?

ajelinski:
Sure, it seems possible. I will work on that.

PR #56, opened 2020-12-01
Don't merge constants for yosys for the yosys conda package.

litghost:
This is because some strings that get binary patched by the conda
installer might overlap with some internal strings used by yosys.

Without disabling merge-constants, the yosys share search path was
changing from "<binary path>/../share/yosys/" to "<binary path>/../share/" as
a result from conda stirng patching.

See https://github.com/SymbiFlow/conda-packages/pull/119


litghost:
@PiotrZierhoffer Please review and merge once ready.

PiotrZierhoffer:
lgtm, merged.

PR #57, opened 2020-12-01
nextpnr-ice40: Fix building on macOS

ajelinski:
The boost packages v1.73 conflict with the clang 4.0 on macOS. It seems
it requires newer version of the compiler.

However, packages such as libraries linked should be placed in a `host`
environment. Also, there's rarely a need to have the same package as
both the `build` and the `host` requirement.

Therefore, to solve conflicts, the packages mentioned in both have been
removed from the `build` environment.


PiotrZierhoffer:
Merging, this is green on branch

PR #61, opened 2020-12-16
Add python-constraint to avoid conda-forge dependencies.

HackerFoo:
Any packages with a `condarc` specifying outside channels will require those channels to be available to install them.

Depending in conda-forge is undesirable because it can conflict with many of the default packages. `pip` cannot be used directly in a conda recipe, so PyPI packages must be converted to Conda packages.

This has caused problems with `vtr-optimized` ([PR](https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1892)), which this PR aims to resolve, but there are other packages using `condarc` as well.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,6 +26,11 @@ jobs:
     dist: xenial
     env:
     - PACKAGE=misc/magic
+  - stage: "No dependencies"
+    os: linux
+    dist: xenial
+    env:
+    - PACKAGE=misc/python-constraint
 
  # EDA Tools - Bitstream Tools
   # Linux
@@ -183,7 +188,7 @@ jobs:
     dist: xenial
     env:
     - PACKAGE=pnr/vtr-gui
-  - stage: "No dependencies"
+  - stage: "Has first level dependencies"
     os: linux
     dist: xenial
     env:
--- a/misc/python-constraint/meta.yaml
+++ b/misc/python-constraint/meta.yaml
@@ -1,0 +1,51 @@
+{% set name = "python-constraint" %}
+{% set version = "1.4.0" %}
+
+package:
+  name: "{{ name|lower }}"
+  version: "{{ version }}"
+
+source:
+  url: "https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.bz2"
+  sha256: 501d6f17afe0032dfc6ea6c0f8acc12e44f992733f00e8538961031ef27ccb8e
+
+build:
+  # number: 201803050325
+  number: {{ environ.get('DATE_NUM') }}
+  # string: 20180305_0325
+  string: {{ environ.get('DATE_STR') }}
+  script_env:
+    - CI
+    - TRAVIS
+    - CONDA_OUT
+    - TEST_PACKAGE
+  script: "{{ PYTHON }} -m pip install . -vv"
+
+requirements:
+  host:
+    - pip
+    - python
+  run:
+    - python
+
+test:
+  imports:
+    - constraint
+    - examples
+    - examples.abc
+    - examples.coins
+    - examples.crosswords
+    - examples.einstein
+    - examples.queens
+    - examples.rooks
+    - examples.studentdesks
+    - examples.sudoku
+    - examples.wordmath
+    - examples.xsum
+
+about:
+  home: "https://github.com/python-constraint/python-constraint"
+  license: BSD
+  license_family: BSD
+  license_file: LICENSE
+  summary: "python-constraint is a module implementing support for handling CSPs (Constraint Solving Problems) over finite domain"
--- a/pnr/vtr-optimized/condarc
+++ b/pnr/vtr-optimized/condarc
@@ -1,2 +1,0 @@
-channels:
-  - conda-forge


PiotrZierhoffer:
Ha, we just discussed adding this, like minutes ago.

LGTM for me, but I'll ask @ajelinski to take a look as well

.travis.yml
```diff
@@ -26,6 +26,11 @@ jobs:
     dist: xenial
     env:
     - PACKAGE=misc/magic
+  - stage: "No dependencies"
+    os: linux
+    dist: xenial
+    env:
+    - PACKAGE=misc/python-constraint
```
PiotrZierhoffer, line 33: this should, probably, go to litex-conda-misc, but since it's a direct dependency of vtr-optimized, I don't think it's a big problem to have it here

.travis.yml
```diff
@@ -183,7 +188,7 @@ jobs:
     dist: xenial
     env:
     - PACKAGE=pnr/vtr-gui
-  - stage: "No dependencies"
+  - stage: "Has first level dependencies"
```
litghost, line 191: I don't think you meant to move every package into first level deps?

litghost: Nevermind

litghost (approved):
LGTM

misc/python-constraint/meta.yaml
```diff
@@ -0,0 +1,51 @@
+{% set name = "python-constraint" %}
```
PiotrZierhoffer, line 1: is this file autogenerated with something like `conda skeleton pypi {packagename}`? I actually never used it before, but if this is the result, then it's pretty good

PiotrZierhoffer (approved):

acomodi:
There should be another way to fix the python-constraint problem. Basically the condarc file in `vtr-optimized` should include also the `defaults` package.

All the dependencies for vtr-optimized will be taken from `defaults` apart from python-constraint which will be taken from conda-forge. Given that python-constraint does not have any other dependency itself, it should not generate any conflicts.

So basically the condarc file would look like the following:
```
channels:
  - defaults
  - conda-forge
```

in the current implementation of the vtr-optimized package, all the various dependencies are taken from conda-forge, hence the conflicts

misc/python-constraint/meta.yaml
HackerFoo: Yeah, but I had to edit a bit, and change from .gz to .bz2.

HackerFoo:
> There should be another way to fix the python-constraint problem. Basically the condarc file in `vtr-optimized` should include also the `defaults` package.
> 
> All the dependencies for vtr-optimized will be taken from `defaults` apart from python-constraint which will be taken from conda-forge. Given that python-constraint does not have any other dependency itself, it should not generate any conflicts.
> 
> So basically the condarc file would look like the following:
> 
> ```
> channels:
>   - defaults
>   - conda-forge
> ```
> 
> in the current implementation of the vtr-optimized package, all the various dependencies are taken from conda-forge, hence the conflicts

@acomodi  I tried this, based on [this tip](https://conda-forge.org/docs/user/tipsandtricks.html), except reversed to avoid conda-forge, but unfortunately, it didn't work. There were still dependencies from conda-forge in `info/about.json`.

HackerFoo:
@ajelinski Any comments?

ajelinski:
Are you sure you tried with channels in this exact order? The order matters and in the link provided `conda-forge` is the most important channel.

I've tested how the recipe is rendered after adding the `defaults` channel on top in `condarc` like @acomodi mentioned. It definitely solves the problem which is caused by requiring libraries that are not available on the `defaults` (ie. `main`) channel:
```
  run:
  - libstdcxx-ng >=9.3.0
  - libgcc-ng >=9.3.0
  - tbb >=2020.2
```
This is because conda-build gets all packages it needs from `conda-forge` and there is gcc/gxx 9.3.0. After making `defaults` the most important channel (the first one in `condarc`) it becomes:
```
  run:
  - libstdcxx-ng >=7.3.0
  - tbb >=2020.3
  - libgcc-ng >=7.3.0
```
because conda-build will use all packages it can from the `defaults` channel. Only those not available will be taken from the `conda-forge` channel.

Those requirements can be satisfied without `conda-forge` so it won't be needed to install `vtr-optimized` after such change.

ajelinski:
Btw. `info/about.json` doesn't matter during dependency resolving. The `depends` key from `info/index.json` inside the package is what matters the most and there you can see this exact problem with requiring `libgcc/libstdcxx >=9.3.0`.

ajelinski:
I have added a branch with the aforementioned `condarc` change and all run requirements of the built package can be satisfied without `conda-forge`: https://travis-ci.com/github/litex-hub/litex-conda-eda/jobs/461371641#L2925

The package built in that CI can be installed with:
```
conda install -c litex-hub/label/travis-fix-vtr-optimized-209189450 vtr-optimized
```

HackerFoo:
@ajelinski Thanks. I'll try you package and close this if it works, unless there is still value in it.

HackerFoo:
@ajelinski It seems to work, so made a PR with your branch: https://github.com/litex-hub/litex-conda-eda/pull/62

HackerFoo:
Superseded by https://github.com/litex-hub/litex-conda-eda/pull/62

PR #62, opened 2020-12-16
vtr-optimized: Give highest priority to the `defaults` channel

HackerFoo:
Such change will make `conda-build` get all required packages from the
`defaults` channel if they only exist there. Only the packages that
aren't available in that channel will be taken from the `conda-forge`
channel (`python-constraint`).

### DIFF
--- a/pnr/vtr-optimized/condarc
+++ b/pnr/vtr-optimized/condarc
@@ -1,2 +1,3 @@
 channels:
+  - defaults
   - conda-forge


HackerFoo:
This is a better fix for https://github.com/litex-hub/litex-conda-eda/pull/61 by @ajelinski

litghost (approved):
Merge if green

PR #64, opened 2021-01-04
Move to ghactions

tjurtsch:


syn/symbiflow-yosys-plugins/meta.yaml
```diff
@@ -1,8 +1,6 @@
-{% set version = '%s_%04i_%s'|format(GIT_DESCRIBE_TAG|replace('v','') or '0.X', GIT_DESCRIBE_NUMBER|int, GIT_DESCRIBE_HASH or 'gUNKNOWN') %}
-
 package:
   name: symbiflow-yosys-plugins
-  version: {{ version }}
+  version: 0.0
```
PiotrZierhoffer: is this change needed?

sim/verilator/run_test.sh
```diff
@@ -3,6 +3,7 @@
 set -e
 set -x
 
+export CONDA_BUILD_SYSROOT=/Users/runner/sdk/MacOSX10.9.sdk
```
PiotrZierhoffer, line 6: is this something we should have on a generic level?

sim/icarus/meta.yaml
```diff
@@ -6,7 +6,7 @@ package:
 
 source:
   git_url: https://github.com/steveicarus/iverilog.git
-  git_rev: master
+  git_rev: 540555fca318f142fbddaac7c93b9dc66c1febdb
```
PiotrZierhoffer, line 9: This is a temporary solution that we'll merge and then revert after the package is built

PiotrZierhoffer (changes requested):

syn/symbiflow-yosys-plugins/meta.yaml
tjurtsch: It's not. Reverted it.

sim/verilator/run_test.sh
tjurtsch: This SDK is downloaded and unpacked in this custom location by download_sdk.sh script. Without pointing to it test won't pass: https://github.com/litex-hub/litex-conda-eda/runs/1662891928?check_suite_focus=true
Wrong SDK is used

tjurtsch: Piotr means generic level as e.g. workflow level. We can't have this env variable defined for whole workflow, because Conda overrides these variables and creates separate env for running tests.

.github/workflows/build-packages.yml
```diff
@@ -0,0 +1,639 @@
+name: build-packages
+on: push
```
ajelinski, line 2: Change to `on: [push, pull_request]`

ajelinski:
LGTM

PR #66, opened 2021-01-12
Fix recipes to only use Conda-based tools

ajelinski:
After trying to build the recipes on a clean Docker container created from the `ubuntu` image it turned out many of them are relying on the tools preinstalled in the Travis and GitHub Actions Operating Systems.

The changes make all of the recipes buildable without requiring any additional build tools installed in the container.


PiotrZierhoffer (approved):
lgtm

PR #68, opened 2021-01-21
surelog/yosys: Allow running with newer Python than used during building

ajelinski:
The change is required for `surelog` and `yosys` to work in the SystemVerilog testing workflow: https://github.com/SymbiFlow/sv-tests because they are used alongside Python 3.8. Python 3.7 is still used to build the packages, the only change is about allowing them to be run with Python >=3.7.

`yosys` requires using more complicated measures to achieve that because it has Python as a `host` requirement. Even though it's also one of the `build` requirements and one of them seems to be redundant, I've checked that building fails without the `host` one.

Therefore `conda_build_config.yaml` had to be used to loosen the version restriction (by default Python has `max_pin: x.x`). I removed the old contents of the file since it certainly had no influence. The currently built `yosys` packages require Python 3.7 to run even though according to the file (together with the `python {{ python }}` run requirement) specified it should be Python 3.6.

The `surelog-uhdm` and `verilator-uhdm` changes are simple fixes because somehow the `run` key got lost during their addition based on the `symbiflow/conda-packages` recipe.


mithro:
@ajelinski Can you just build everything for both Python 3.7 and Python 3.8?

ajelinski:
@mithro You mean all of the `litex-conda-*` packages?

PiotrZierhoffer:
@mithro do you mean building two flavors of each package? Or relaxing the dependencies like here?

mithro:
> @mithro do you mean building two flavors of each package? Or relaxing the dependencies like here?

Keep strict dependencies and building two flavors of each package. 

> @mithro You mean all of the `litex-conda-*` packages?

Yes.

ajelinski:
@mithro This can be done but I don't think it should be applied automatically to all of the packages. I mean, Python version doesn't always matter as it is not always required to build the package and sometimes it's only used to run some build scripts. Therefore many CI jobs would be redundant.

I've just modified the PR change since maybe you're right that it's better to have separate packages for Python 3.7 and 3.8. `surelog` and `yosys` definitely need some sort of a change to work with `sv-tests` so it'd be nice to have at least these two packages built with Python 3.8 for now.

mithro:
@ajelinski -- All packages which require Python should build for Python 3.7, Python 3.8, Python 3.9

The reason I was suggesting everything is there is some subtle complexity around the dependencies of Python itself can conflict with dependencies of the other packages.

.github/workflows/build-packages.yml
```diff
@@ -331,7 +345,8 @@ jobs:
   #26
   symbiyosys-linux:
```
mithro, line 346: You should probably build SymbiYosys against both these.

ajelinski: Done.

ajelinski:
I'll merge the PR since it's been successfully tested already apart from the `symbiyosys` change that won't be tested because that job is currently skipped as it has been already failing before.

The `tree-sitting-verilog` started failing from the unrelated reasons so I've skipped it and created an issue for it: #73. I've also created an issue to add all the missing Python 3.8/3.9 builds: #72.

PR #75, opened 2021-02-04
Build uhdm-integration packages for python 3.7 and 3.8

lpawelcz:
This PR introduces additional `uhdm-integration` packages with python 3.8 dependency. This is required in SystemVerilog testing workflow (encountering the same problems as in https://github.com/litex-hub/litex-conda-eda/pull/68)

### DIFF
--- a/.github/workflows/build-packages.yml
+++ b/.github/workflows/build-packages.yml
@@ -3,7 +3,7 @@ on: [push, pull_request]
 env:
   ANACONDA_TOKEN: ${{ secrets.ANACONDA_TOKEN }}
   ANACONDA_USER: ${{ secrets.ANACONDA_USER }}
-  NUM_OF_JOBS: 66
+  NUM_OF_JOBS: 68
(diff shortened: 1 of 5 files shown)


ajelinski (changes requested):
Now there are two `uhdm-integration-verilator` and two `uhdm-integration-yosys` packages built that'll have the same version string. They need to have unique build strings. I'd add the same `_py3X` suffix as in `surelog-uhdm` and `yosys-uhdm`.

lpawelcz:
changes in v2:
* changed preparing 2 versions of each `uhdm-integration` metapackages  to one version that will reference `surelog-uhdm` and `yosys-uhdm` in python 3.8 and 3.7

ajelinski (approved):
LGTM now. It will allow using `uhdm-integration-*` metapackages with Python 3.8. The metapackages itself doesn't need separate builds based on the Python version since they don't tie any specific versions of packages.

PR #76, opened 2021-02-08
Fixes for uhdm-yosys package

kamilrakoczy:
This PR fixes encouraged problems with ``uhdm-yosys`` package in ``fpga-tool-perf`` :

- includes whole yosys install in package (not only binary yosys file)
- uses conda compiler to compile yosys (and sets gcc/gxx aliases to conda compiler, as ``abc`` is using them directly, this fixes the problem with wrong path to yosys - share directory after compilation)
- uses ``PROGRAM_PREFIX`` to add prefix to the output files - this way we won't conflict with other yosys installed in the system, but it changes binary name to ``uhdm-yosys`` from ``yosys-uhdm`` (this means we need to adapt all repositories that is already using this package after merge.).

### DIFF
--- a/.github/workflows/build-packages.yml
+++ b/.github/workflows/build-packages.yml
@@ -788,7 +788,6 @@ jobs:
     runs-on: "ubuntu-16.04"
     env:
       PACKAGE: "syn/yosys-uhdm"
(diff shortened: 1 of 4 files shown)


ajelinski (approved):
LGTM

PR #81, opened 2021-02-25
Hdl migration

PiotrZierhoffer:
Updating repo references, sumbodules etc


PiotrZierhoffer:
We're not green because of some network issues. I'll merge anyway, as the same code was tested on litex-hub

PR #82, opened 2021-02-25
Correct metapackages

ajelinski:
This PR makes all metapackages always tied precisely to the packages built in the same CI workflow. This is required to make locking such metapackages possible and making sense.

However, there is a slight downside. Building a metapackage now will never succeed in cross-repository PRs. There's no Anaconda token in such workflows so the dependencies aren't uploaded after building and Conda won't be able to find them building the metapackage.

Building these recipes was tested locally. Jobs building metapackages are expected to fail in this PR if everything works well.

### DIFF
--- a/.github/workflows/build-packages.yml
+++ b/.github/workflows/build-packages.yml
@@ -18,7 +18,7 @@ on:
 env:
   ANACONDA_TOKEN: ${{ secrets.ANACONDA_TOKEN }}
   ANACONDA_USER: ${{ secrets.ANACONDA_USER }}
-  NUM_OF_JOBS: 70
+  NUM_OF_JOBS: 71
 defaults:
   run:
     shell: bash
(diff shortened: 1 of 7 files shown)


.github/workflows/build-packages.yml
```diff
@@ -10,7 +10,7 @@ on:
 env:
   ANACONDA_TOKEN: ${{ secrets.ANACONDA_TOKEN }}
   ANACONDA_USER: ${{ secrets.ANACONDA_USER }}
-  NUM_OF_JOBS: 69
+  NUM_OF_JOBS: 71
```
mithro, line 13: Can you log an issue to remove this hard coded number?

.github/workflows/build-packages.yml
```diff
@@ -830,27 +830,59 @@ jobs:
       - uses: hdl/conda-ci@master
 
   #59
-  uhdm-integration-verilator-linux:
+  uhdm-integration-verilator-linux-py37:
```
mithro, line 833: I feel like this style be better done with a matrix?

mithro (approved):
I /think/ this looks good. If the CI says so, please go ahead and merge.

.github/workflows/build-packages.yml
ajelinski: It was done this way because you can't set name of a job and its required jobs (`needs`) based on a matrix variable.

Therefore using a matrix some package using Python 3.8 can fail because its required job has a matrix to build packages using multiple versions of Python and any of them failed, even when Python 3.8-based package was successfully built. I'm not sure if the slightly better style is worth it.

The workflow file is huge and complex nevertheless. I think the only real improvement would be to create `build-packages.yml` with a script based on some compact job definitions file.

ajelinski:
Tested metapackages in CI on `hdl` branch since now they will always be failing in cross-repository PRs (which is absolutely correct): https://github.com/hdl/conda-eda/actions/runs/636011644

I've checked one of the metapackages, `uhdm-integration-yosys-0.5_8397_ge1d4ed07a-20210309_140417_py38` and its dependencies are exactly as they should be:
* `surelog-uhdm 0.0_2133_ge0b10a630 20210309_140417_py38`,
* `yosys-uhdm 0.5_8397_ge1d4ed07a 20210309_140417_py38`.

ajelinski:
I've added a commit to skip metapackage-building jobs if there's no ANACONDA_TOKEN, so mainly in cross-repository PRs such as this one. This will allow having CI green in such PRs. I can see the `quicklogic-vtr` has been skipped already so it works.

PR #83, opened 2021-03-02
Fix runtime dependencies for Verilator

RRozak:
I have added the following required runtime dependiencies to Verilator: cxx, bison, flex and ncurses

### DIFF
--- a/sim/verilator/meta.yaml
+++ b/sim/verilator/meta.yaml
@@ -30,6 +30,11 @@ requirements:
     - bison
     - flex
     - zlib
+  run:
+    - {{ compiler('cxx') }}
+    - bison
+    - flex
+    - ncurses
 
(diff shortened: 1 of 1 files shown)


tgorochowik:
These dependencies are copied from the `uhdm-verilator` package.

This resolves the issue with a constant number of new failures for Verilator in the PR comment job (e.g. here: https://github.com/SymbiFlow/sv-tests/pull/1368#issuecomment-787031986)

The issue is that the comparison uses a build with all the packages installed at once as the baseline (so we do get the compiler from a dependency from another package there) and compare that against jobs which only install individual packages (thus the lack of dependencies for Verilator)

PiotrZierhoffer (approved):
The fail is http-related, so lgtm

PR #85, opened 2021-03-05
Add periodic builds

tgorochowik:
The packages are usually built from their default branches (the version
is not frozen).

For now we only had builds triggered by changes in the repository with
the packages.

The tool repositories change independently of the conda package
repository.

Adding a periodic build allows us to keep the packages up to date
without the need to update this repo and without triggering anything
manually.

### DIFF
--- a/.github/workflows/build-packages.yml
+++ b/.github/workflows/build-packages.yml
@@ -7,6 +7,8 @@ on:
     branches: [ master ]
   pull_request:
   workflow_dispatch:
+  schedule:
+    - cron: '0 23 * * *' # run daily at 23:00 (UTC)
 env:
   ANACONDA_TOKEN: ${{ secrets.ANACONDA_TOKEN }}
   ANACONDA_USER: ${{ secrets.ANACONDA_USER }}


.github/workflows/build-packages.yml
```diff
@@ -7,6 +7,8 @@ on:
     branches: [ master ]
   pull_request:
   workflow_dispatch:
+  schedule:
+    - cron:  '0 1 * * *'
```
mithro, line 11: I don't know the cron syntax off the top of my head, so it would be good to include the extra comment about the frequency you are targeting here.

mithro (approved):
Good idea!

Just one comment.

PiotrZierhoffer:
Can you please add it to conda-compilers, conda-misc and conda-prog as well?

PiotrZierhoffer (approved):

.github/workflows/build-packages.yml
tgorochowik: Added

umarcor:
@tgorochowik CI limits are shared in the organization. Open source projects/organizations have "unlimited" CI resources, but there is a maximum number of 20 concurrent workflows and/or jobs: https://docs.github.com/es/actions/reference/usage-limits-billing-and-administration#usage-limits.

Some months ago, excessive jobs would be weirdly hidden/frozen. Nowadays, those are scheduled, as explained in the docs. Therefore, this is not a critical problem. However, I suggest scheduling these repositories at different hours, for allowing other repos/PRs/schedules in the organization to be processed in-between.

tgorochowik:
@umarcor that's a valid point, thanks, I will update the PRs to make sure the builds don't overlap.

tgorochowik:
The schedule is now as follows:
23:00 - conda-eda
01:00 - conda-prog
03:00 - conda-compilers
05:00 - conda-misc

(UTC).

The times to run are completely arbitrarily and can be changed.

umarcor (approved):

PR #86, opened 2021-03-08
Fix antmicro-yosys installation

RRozak:
The build scripts in the previous version were installing yosys binaries with the original names. It was causing conflicts with the original yosys package.
Now all the binaries from this package will use the "antmicro-" prefix, so there won't be any conflicts.

### DIFF
--- a/syn/antmicro-yosys/build.sh
+++ b/syn/antmicro-yosys/build.sh
@@ -9,7 +9,14 @@ cd yosys
 
 make V=1 -j$CPU_COUNT
 make install V=1 -j$CPU_COUNT
-cp yosys "$PREFIX/bin/antmicro-yosys"
-cp yosys-config "$PREFIX/bin/yosys-config"
+
+pushd $PREFIX/bin
+
+for filename in `ls yosys*`;
+do
+    mv $filename "antmicro-"$filename
+done
+
+popd
 
 $PREFIX/bin/antmicro-yosys -V


umarcor (commented):
Making the binary names of forks unique is a good idea. However, this PR might break other things as is...

- There is a feature in Yosys' Makefile for doing this: https://github.com/YosysHQ/yosys/blob/master/Makefile#L52. Setting `PROGRAM_PREFIX` should be equivalent to this PR, and to L15-L16. Should it not be, it should be reported upstream.
- This program prefix customization should be applied to the other forks too (quicklogic and symbiflow).
- The scripts/tools that consume these binaries might need to be updated accordingly.

BTW, @mithro what is the status of those three forks? Are all of them independent (not compatible) and required?

tgorochowik:
@umarcor thanks for your comments!

The prefix is important in this particular package because this tool is used in [sv-tests](https://github.com/SymbiFlow/sv-tests). This package is installed together with vanilla yosys in a single run (kokoro).

Note that this change does not really add anything new, the issue it fixes is that earlier we used `cp` in addition to the regular `make install` call. Therefore we had yosys installed both as `yosys` and `antmicro-yosys`. Now we only rename the installed binaries.

We probably should not change the prefix for the quicklogic and symbiflow forks, or at least we should not do this without notifying all the people that already use it.

This PR does not break anything, but fixes co-existing installation of yosys and antmicro-yosys (which only worked if the tools were installed in the correct order, now the order shouldn't matter).

RRozak:
Thanks, @umarcor 
Setting `PROGRAM_PREFIX` variable does the same. I will correct the diff.

umarcor:
> We probably should not change the prefix for the quicklogic and symbiflow forks, or at least we should not do this without notifying all the people that already use it.

@tgorochowik, is there any brief description of this anywhere? For instance:

"*Note that antmicro and upstream yosys installations can coexist. However, the quicklogic and symbiflow versions need to be installed as a replacement.*"

umarcor (approved):

tgorochowik:
> @tgorochowik, is there any brief description of this anywhere? (..)

I don't think so. It probably makes sense to add something like this but I don't think it should be a part of this PR. Would you mind creating an issue for that?

umarcor:
@tgorochowik, I agree that such an update is unrelated to this PR.

Related to hdl/conda-eda#87, there is currently no documentation entrypoint for users. Hence, I will delay the clarification until that one is defined.

ajelinski:
Unfortunately `PROGRAM_PREFIX` changes not only `yosys` name but also at least `yosys-config`, which is why `antmicro-yosys-plugins` fails currently: https://github.com/hdl/conda-eda/issues/91 .

tgorochowik:
The change was by design (we need to rename everything if the tools are to co-exist). #92 attempts to fix the issue introduced by this.

PR #87, opened 2021-03-08
README: Update the local building section

ajelinski:
Apart from the differences caused by a different repository name/URL and the package built by default (`symbiflow-yosys`) only the recipe-dependent environment variables part is different from the local building section merged in https://github.com/hdl/conda-compilers/pull/15 :
```
Currently required additional environment variables are:
* `LIBFFI_VERSION` (by: `syn/symbiflow-yosys`) – must contain a valid version
  of `libffi` Conda package, e.g., `3.3`.
```

I will merge it once the `tuttest` CI is green.

### DIFF
--- a/.github/workflows/build-packages.yml
+++ b/.github/workflows/build-packages.yml
@@ -5,7 +5,13 @@ on:
     # the same commit shouldn't run simultaneously because they're overwriting
     # each other's packages on Anaconda.
     branches: [ master ]
+    paths-ignore:
+      - '.github/workflows/tuttest.yml'
+      - 'README.md'
   pull_request:
+    paths-ignore:
+      - '.github/workflows/tuttest.yml'
+      - 'README.md'
   workflow_dispatch:
 env:
   ANACONDA_TOKEN: ${{ secrets.ANACONDA_TOKEN }}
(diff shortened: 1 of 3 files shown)


umarcor (commented):
It seems that the same long README is being proposed in multiple repositories. I would propose to avoid doing so. It increases the maintenance burden and it is misleading for users, because they need to read everything in order to know whether any different content is explained.

Instead, I would propose to have a single main documentation point. That can be the README of a single repo (this one, or hdl/conda-ci or that one renamed to hdl/conda) or an Sphinx/asciidoctor site (say hdl.github.io/conda). Other readmes should have a brief description and then point to the main documentation/entrypoint.

ajelinski:
@umarcor I get your point, it's definitely a valid one but the `conda-compilers` README was created with an intention that all repositories containing recipes have their own local building sections. In case we want to have a single place with that information the already-merged README of `conda-compilers` will need to be changed.

We can work on such a unification but I don't think that it needs to postpone this change. The READMEs need further changes anyway, especially after the `hdl` migration. Maybe then it can be worked out how to have a single README.

umarcor:
IMHO, it does not harm to delay these changes a few days, even considering the naming migration. Moreover, it is legit to update the README in `conda-compilers` (even if it was merged recently), because that allows reducing noise-to-signal. Nonetheless, I won't oppose to these PRs, should that be how you want to handle it.

Issue #84, opened 2021-03-04
Ppackages should have fixed runtime dependencies 


kgugala:
Currently packages define runtime dependencies without specifying version (so the latest one is used). We should force a specific version where possible.

An example of such dependency can be seen here https://github.com/hdl/conda-eda/blob/master/pnr/vtr/meta.yaml#L38.
This particular one hit us in the `symbiflow-examples` repo - newer tbb package does not provide `libtbb.so.2` required by VPR, but rather `libtbb.so.12`.

PR #90, opened 2021-03-09
Don't use incompatible v2021 `tbb` and `tbb-devel` packages

ajelinski:
Related issue: #84

The `tbb-2021` and `tbb-devel-2021` available on `conda-forge` aren't
backward compatible. In fact, they will never be backward compatible
again: https://software.intel.com/content/www/us/en/develop/articles/intel-oneapi-threading-building-blocks-release-notes.html

The `tbb-2021` can't be used to run a package built with `tbb-2020`
because it even has a main library renamed from `libtbb.so.2` to
`libtbb.so.12`.

The new packages can't be currently built with TBBv2021 because CMake
isn't able to find it. It seems `include/tbb/tbb_stddef.h` is used
during the lookup but that file has been removed from `tbb-devel`
package. Also, currently `v2021` is only available from `conda-forge`.

### DIFF
--- a/pnr/odin_II/meta.yaml
+++ b/pnr/odin_II/meta.yaml
@@ -34,9 +34,11 @@ requirements:
     - xorg-libxcb
     - xorg-libxext
     - xorg-libxft
-    - tbb
-    - tbb-devel
+    - tbb <2021.0.0a0
+    - tbb-devel <2021.0.0a0
     - gtk3
+  run:
+    - tbb <2021.0.0a0
 
 about:
   home: http://verilogtorouting.org/
--- a/.github/workflows/build-packages.yml
+++ b/.github/workflows/build-packages.yml
@@ -916,6 +916,7 @@ jobs:
     env:
       PACKAGE: "sim/antmicro-yosys-complete"
       OS_NAME: "linux"
+      SKIP: "true"  # See https://github.com/hdl/conda-eda/issues/91
     steps:
       - uses: actions/checkout@v2
         with:
@@ -929,6 +930,7 @@ jobs:
     env:
       PACKAGE: "syn/antmicro-yosys-plugins"
       OS_NAME: "linux"
+      SKIP: "true"  # See https://github.com/hdl/conda-eda/issues/91
     steps:
       - uses: actions/checkout@v2
         with:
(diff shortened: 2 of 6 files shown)


ajelinski:
This PR should make master green again.

ajelinski:
I see there's additional problem with `antmicro-yosys-plugins` that won't make master green. Added an issue #91 and a commit to skip these builds until the issue is solved.

Unfortunately I've also noticed tha skipping metapackage builds when `ANACONDA_TOKEN` isn't available: https://github.com/hdl/conda-eda/commit/9fd2473a136927a3507a2c9b61ff5b646a425b97 isn't working properly because these builds were skipped on master CI too. I used some other, even better if it's correct, option to skip these builds in cross-repository PRs.

@umarcor do you think this will work as expected now? https://github.com/hdl/conda-eda/pull/90/commits/c90de4320809fc47321eb813a54a3818e377efed

ajelinski:
I've tested whether the job skipping in cross-repository PRs works properly now and it does: https://github.com/hdl/conda-eda/pull/93 . Since jobs with that addition were successfully run here, i.e., in intra-repository PR, it seems that skipping works only when there's no `ANACONDA_TOKEN` indeed.

I'll rebase the PR and merge if green. There was a PR merged in the meantime that seemed to solve `antmicro-yosys-plugins` but it only worked because it used `antmicro-yosys` from Anaconda. Therefore the commit disabling that build is left intact.

PiotrZierhoffer:
Looks fine. We'll address the failing packages separately, but we need the CI green

pnr/odin_II/meta.yaml
```diff
@@ -34,9 +34,11 @@ requirements:
     - xorg-libxcb
     - xorg-libxext
     - xorg-libxft
-    - tbb
-    - tbb-devel
+    - tbb <2021.0.0a0
```
litghost, line 37: This seems like it will eventually bite us if another package requires the new version of tbb?  Why can the new package not but used?

litghost: I think we want to be something like https://docs.conda.io/projects/conda-build/en/latest/resources/variants.html#pinning-at-the-recipe-level (e.g. `pin_compatible`)

PR #92, opened 2021-03-10
Fix antmicro-yosys-plugins

### LINKED ISSUE #91: Fix antmicro-yosys-plugins after renaming yosys-config
`yosys-config` was renamed in https://github.com/hdl/conda-eda/commit/d646059e71222f7f73836b36dbd285a13cca12ac and it seems to be the root of the `antmicro-yosys-plugins` failure: https://github.com/hdl/conda-eda/runs/2070486993

I'll add a commit in https://github.com/hdl/conda-eda/pull/90 to skip `antmicro-yosys-plugins-linux` and `antmicro-yosys-complete-linux` jobs until this is fixed to make master CI green again.

tgorochowik:
Attempt to fix #91

### DIFF
--- a/syn/antmicro-yosys-plugins/antmicro-yosys.patch
+++ b/syn/antmicro-yosys-plugins/antmicro-yosys.patch
@@ -1,0 +1,69 @@
+diff --git a/.github/workflows/setup.sh b/.github/workflows/setup.sh
+index abe6e8d..17eb1b2 100755
+--- a/.github/workflows/setup.sh
++++ b/.github/workflows/setup.sh
+@@ -43,9 +43,9 @@ start_section Install-Yosys
+         cd yosys
+         PREFIX=$HOME/.local-bin make -j$(nproc)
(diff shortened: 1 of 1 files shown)


ajelinski:
LGTM. `antmicro-yosys-plugins` is fixed indeed.

ajelinski:
OK I've just realized we don't know if the patch really fixes the issue. This was a cross-repository PR which means, it didn't use `antmicro-yosys` built in CI but rather the one from `litex-hub` channel labeled `main`. That package still has binaries not prefixed with `antmicro-`. In addition, I didn't notice that the patch wasn't added to `meta.yaml` so it wasn't really applied.

Right now I'm testing whether adding a patch to `meta.yaml` is enough to make it right: https://github.com/hdl/conda-eda/actions/runs/639348084

PiotrZierhoffer:
Had to restart the CI after merging - it failed because it didn't find available macos runners, I think

PR #94, opened 2021-03-11
Create symlinks before building antmicro yosys plugins

tgorochowik:
Having symlinks for the purpose of building the packages should be
enough and will not require us to maintain the patch.


ajelinski:
I added a commit that reenables `antmicro-yosys-plugins/-complete`. Let's see if they will be successfully built.

ajelinski:
LGTM

PR #100, opened 2021-03-31
symbiflow-yosys-plugins: Add patch restoring prev. flag setting in make

ajelinski:
Flag setting introduced in https://github.com/SymbiFlow/yosys-symbiflow-plugins/commit/f02851af094f09d641972b7b3572014848e96615
results in the compiler flags not being set. That's because Conda's
compiler packages always set their own flags in the `conda-build`
working environment.

The patch simply reverses that commit.


kgugala (approved):
LGTM - CI failures are unrelated to this PR, `symbiflow-yosys-plugins` build is green again

PR #101, opened 2021-04-02
yosys-uhdm fix: Use some build requirements as the host ones instead

ajelinski:
Not only it solves build conflicts, which roots I'm unable to pinpoint,
but it's also more appropriate than the previous requirements
arrangement.


ajelinski:
Testing confirmed the fix works. `verible` and `slang` have their own problems.

PR #103, opened 2021-04-09
slang: verible: Use gcc-9

ajelinski:
GitHub Actions virtual environments doesn't have gcc-8 preinstalled
anymore: https://github.com/actions/virtual-environments/issues/2950

### DIFF
--- a/.github/workflows/build-packages.yml
+++ b/.github/workflows/build-packages.yml
@@ -681,7 +681,7 @@ jobs:
     runs-on: "ubuntu-16.04"
     env:
       PACKAGE: "sv-front/verible"
-      USE_SYSTEM_GCC_VERSION: "8"
+      USE_SYSTEM_GCC_VERSION: "9"
       OS_NAME: "linux"
     steps:
       - uses: actions/checkout@v2
@@ -720,7 +720,7 @@ jobs:
     env:
       PACKAGE: "sv-front/slang"
       OS_NAME: "linux"
-      USE_SYSTEM_GCC_VERSION: "8"
+      USE_SYSTEM_GCC_VERSION: "9"
     steps:
       - uses: actions/checkout@v2
         with:


PiotrZierhoffer (approved):
Lgtm, merge on green

tgorochowik:
Can we use gcc-10 instead? It should be available:

https://github.com/actions/virtual-environments/blob/fbcc41275becc2c37dc3bfb9c98e1ab879b0b125/images/linux/toolsets/toolset-2004.json#L249-L253

mithro:
You should really be using the Compiler provided in the conda environment. Using the system provided compiler is generally bad idea.

ajelinski:
@tgorochowik This would also require updating the image we build on. Currently used Ubuntu 16.04 for all Linux builds doesn't have gcc-10: https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu1604-README.md

@mithro AFAIR these two were impossible to build with a compiler from Conda and at least `slang` is a static binary. Still, this fix is just an adjustment to the GitHub Actions changes. These two packages have been built with OS-based gcc already.

mithro:
@ajelinski - Can we make sure that is a GitHub issue which is tracking that this problem exists and needs to be fixed in the future?

ajelinski:
@mithro This seems to be a conscious change in the OS image. I can see they are constantly making software changes in the images: https://github.com/actions/virtual-environments/commits/main/images/linux/Ubuntu1604-README.md

umarcor:
My proposal would be to use one single docker image for having an stable build environment, regardless of the host. The purpose of such image is not for users to consume it, but just to use it in CI as a homogeneization layer on top of whichever VM (virtual environment) the CI provider offers. I think that the dockerfile that builds and the CI that updates the image would fit in hdl/conda-ci. We might then consider providing a ready-to-use `hdlc/conda` container image in hdl/containers. But that is a different discussion.

However, I agree with Tim that, in the mid term, it'd be better to have it done regardless of the compiler on the system.

mithro:
> @mithro This seems to be a conscious change in the OS image. I can see they are constantly making software changes in the images: https://github.com/actions/virtual-environments/commits/main/images/linux/Ubuntu1604-README.md

I meant a GitHub issue about the fact the conda package is using the system compiler package.

ajelinski:
@mithro Sorry, clearly I misunderstood you. I've added such an issue: https://github.com/hdl/conda-eda/issues/104

@umarcor I think it's a good idea, maybe you could add a GH issue?

I'm merging the PR since it fixes failures and the doubts aren't about the change itself.

umarcor:
@ajelinski I have this pending in my notifications/ToDo: https://github.com/hdl/conda-ci/pull/13#issuecomment-764915049. I just exposed the idea here for others to be aware.

PR #105, opened 2021-04-14
Fix nextpnr interchange installed device

acomodi:
The installed devices do miss patches information, relevant for constraints and LUTs definition.

This PR installs the correct devices and also removes the symbiflow-yosys run dependency package which should not be needed.

### DIFF
--- a/pnr/nextpnr/fpga_interchange/build.sh
+++ b/pnr/nextpnr/fpga_interchange/build.sh
@@ -36,10 +36,13 @@ mkdir -p $CHIPDB_DIR
 mkdir -p $DEVICES_DIR
 
 # Compute chipdbs for nextpnr-fpga_interchange
+# TODO: change the <device>.device names as soon as more architectures are added.
+#       In fact, the `constraints-luts` nomenclature corresponds
+#       to the patches specifically applied to xc7 devices only.
 for device in $DEVICES; do
 	make chipdb-${device}-bin -j${CPU_COUNT}
-	cp `find -iname "chipdb-${device}.bin"` $CHIPDB_DIR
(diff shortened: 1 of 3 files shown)


pnr/nextpnr/fpga_interchange/build.sh
```diff
@@ -36,10 +36,13 @@ mkdir -p $CHIPDB_DIR
 mkdir -p $DEVICES_DIR
 
 # Compute chipdbs for nextpnr-fpga_interchange
+# FIXME: need to change the <device>.device names as soon as more architectures
```
PiotrZierhoffer, line 39: @acomodi do you track this in an issue somewhere?

PiotrZierhoffer (approved):

pnr/nextpnr/fpga_interchange/build.sh
acomodi: Hmm not really, it's not really an issue, probably this should be a TODO rather than a FIXME

ajelinski (approved):
If you're going to change that `FIXME` @acomodi also please unify the tags in commit messages.

acomodi:
> If you're going to change that `FIXME` @acomodi also please unify the tags in commit messages.

Done

ajelinski:
@acomodi I missed the fact that if the package doesn't require `symbiflow-yosys` anymore then it doesn't need to wait for it in CI. So remove also this line from the workflow file with the `requirements/run` change: https://github.com/hdl/conda-eda/blob/277f77b52db710cfcfed5b9522cb09597a5867c1/.github/workflows/build-packages.yml#L367

After that I will merge the PR as none of these latest changes can cause a CI failure.

PR #106, opened 2021-04-28
Remove uhdm-integration packages

kamilrakoczy:
This PR removes uhdm-integration-* packages as they were using not compatible version of Surelog (different version of Surelog was used to compile yosys/verilator and different version of Surelog was downloaded as part of uhdm-integration package).

Now Surelog is part of verilator-uhdm and yosys-uhdm packages directly.


mithro (commented):
Just wanted to check you can still install `verilator-uhdm` and `yosys-uhdm` at the same time?

mithro:
It seems like surelog should be a package that both these packages depend on?

kamilrakoczy:
No, you shouldn't install ``verilator-uhdm`` and ``yosys-uhdm`` in the same environment (but it is possible, it will just override ``surelog`` version).

The problem is, both ``verilator-uhdm`` and ``yosys-uhdm`` can be compiled with different version of Surelog (not necessary current master revision) and the same version of the Surelog needs to be provided in the tool conda package, otherwise corresponding tool will ``Segmentation fault`` on reading UHDM file.

We can't depend on the same version, e.g. master or currently built version in the CI (as it was in ``uhdm-integration-*`` packages) of the Surelog in both of these packages as it will result in above problem.

sim/verilator-uhdm/meta.yaml
```diff
@@ -11,29 +13,31 @@ source:
...
 
 requirements:
   build:
-    - {{ compiler('c') }}
-    - {{ compiler('cxx') }}
-    - python {{ python }}
+    # These version restrictions make the dependencies resolvable with
+    # packages from the `defaults` channel (libgcc-ng/libstdcxx-ng).
+    - {{ compiler('c') }} <9
+    - {{ compiler('cxx') }} <9
```
ajelinski, line 25: Drop the comment and you can also remove the restrictions. I can see that the package built requires `libunwind` to be installed so it won't work without `conda-forge` anyway.

ajelinski: Also, please remove `unused` from the last commit title. It definitely was used but if the package requires `conda-forge` to be installed then it shouldn't be there.

ajelinski (commented):
Please fix the tag used in this commit title to unify it with other commit tags (`verilator-uhdm`): https://github.com/hdl/conda-eda/pull/106/commits/2024b0c25f191ca5a5800b4e79dc47fc2569b1d2

LGTM apart from these minor issues.

sim/verilator-uhdm/meta.yaml
```diff
@@ -11,29 +13,29 @@ source:
...
     - libuuid
+    - pkg-config
+  host:
+    - python {{ python_version }}
     - gperftools
-    - flex
     - libunwind
     - swig
   run:
-    - {{ compiler('cxx') }}
```
ajelinski, line 32: Not sure about bison or flex but don't you need a C++ compiler to use verilator?

kamilrakoczy: True, I've added them.

ajelinski (approved):

PR #110, opened 2021-06-14
vtr-optimized: remove conda-forge

acomodi:
Fixes the vtr-optimized package build

### DIFF
--- a/pnr/vtr-optimized/build.sh
+++ b/pnr/vtr-optimized/build.sh
@@ -36,6 +36,10 @@ PACKAGE[architecture]=symbiflow-xc7a50t_test-latest
 
 BUILD_ROOT=$PWD
 
+# python-constraint requires to use the conda-forge channel.
+# Given that it is only used to build this package, install it via the GH repository.
+python3 -m pip install git+https://github.com/python-constraint/python-constraint.git
+
 # Executables needed for PGO
 PGO_TARGETS="vpr genfasm"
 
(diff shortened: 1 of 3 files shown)


mithro (commented):
Does `requirements:` support specifying things in `pip`? It doesn't look like it, but it would be worth checking;
 - https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#requirements-section
 - https://conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html#create-env-file-manually

acomodi:
@mithro I have looked a bit at the docs and it does indeed seem that is not possible to install custom pip packages through the `requirements` except by getting them through conda.

mithro:
> @mithro I have looked a bit at the docs and it does indeed seem that is not possible to install custom pip packages through the `requirements` except by getting them through conda.

We should probably log a bug about that on upstream conda. Go ahead and merge otherwise.

mithro (approved):
LGTM

pnr/vtr-optimized/build.sh
```diff
@@ -36,6 +36,10 @@ PACKAGE[architecture]=symbiflow-xc7a50t_test-latest
 
 BUILD_ROOT=$PWD
 
+# python-constraint requires to use the conda-forge channel.
+# Given that it is only used to build this package, install it via the GH repository.
+python3 -m pip install git+https://github.com/python-constraint/python-constraint.git
```
PiotrZierhoffer, line 41: we tend to use `$PYTHON`, can you please change it?

PiotrZierhoffer (approved):

PR #109, opened 2021-06-11
nextpnr: interchange: install also BBASM tool

acomodi:
BBASM is required to produce the binary version of a `.bba` chip information database of nextpnr.

Signed-off-by: Alessandro Comodi <acomodi@antmicro.com>

PR #112, opened 2021-06-15
nextpnr: interchange remove device generation

acomodi:
This PR is an alternative to https://github.com/hdl/conda-eda/pull/109. Given that all the device databases cannot be stored in conda and RAM is limited to build big devices, it is unfeasible to have them generated in here.

Instead, the devices should be generated and stored elsewhere. This would make this package only generating the bbasm and nextpnr-fpga_interchange binaries.


ajelinski (approved):
LGTM

PR #113, opened 2021-06-21
antmicro-yosys: remove conda forge

acomodi:


PiotrZierhoffer:
antmicro-yosys built correctly, I'll merge right away

PR #114, opened 2021-06-21
Fix Arachne compilation failed by newest gcc

PiotrZierhoffer:
This sets the compiler version to be old enough not to complain about
some missing optimisation capabilities. Arachne isn't actively developed
anyway.


PiotrZierhoffer:
Arachne build passed, and it's not a dep for anything, so i merge

PR #116, opened 2021-06-23
bit: add prjoxide

acomodi:


.github/workflows/build-packages.yml
```diff
@@ -60,6 +60,18 @@ jobs:
           submodules: 'true'
       - uses: hdl/conda-ci@master
 
+  #73
```
PiotrZierhoffer, line 63: you have to adjust the num of jobs variable

bit/prjoxide/build.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+set -e
+set -x
+
+curl --proto '=https' -sSf https://sh.rustup.rs | sh -s -- -y
```
PiotrZierhoffer, line 6: Are we sure it's installed within the env? Are there any switches to ensure that, maybe? I assume we do it this way not to use conda-forge?

.github/workflows/build-packages.yml
acomodi: Hmm, the current number is 69, but the last job # is 73 (including prjoxide), should the variable be adjusted to 73 or just incremented to 70?

bit/prjoxide/build.sh
acomodi: Hmm, actually I can try with conda-forge and see if there are compatibility issues with the only other 2 capnproto packages

.github/workflows/build-packages.yml
PiotrZierhoffer: 73

acomodi: Ok, done

bit/prjoxide/build.sh
acomodi: I have checked and using conda-forge generates conflicts

.github/workflows/build-packages.yml
acomodi: Apparently, there are three jobs missing, as https://github.com/hdl/conda-eda/actions/runs/964458839 was hanging forever. I guess this was because the master build was waiting for those three inexistent jobs to finish. I put it back to 70

ajelinski:
Please remove the condarc file in a package. `condarc` with only defaults basically changes nothing.

PR #117, opened 2021-07-14
nextpnr: fix python version and dependencies

acomodi:
The python version should be consistent across all the different nextpnr
versions.

In addition, nextpnr should not require yosys to be built, neither to
run


ajelinski (changes requested):
Please organize and describe the commits more accurately. Add the removal of the remaining `symbiflow-yosys` dependency (your last commit) to the original @acomodi commit (amend it). Remove Python changes from his commit and adjust the commit title/description that it only cleans Yosys deps.

The Python and Boost changes can go together in a commit but you can also separate them. Start the commit titles with the packages being modified. I think `nextpnr: prjtrellis: Unify Python and Boost versions; bump Boost to 1.73` would nicely state what the commit does.

bit/prjtrellis/meta.yaml
```diff
@@ -1,7 +1,9 @@
 # Use `conda-build-prepare` before building for a better version string.
 {% set version = '%s_%04i_%s'|format(GIT_DESCRIBE_TAG|replace('v', '') or '0.X', GIT_DESCRIBE_NUMBER|int, GIT_DESCRIBE_HASH or 'gUNKNOWN') %}
 # cmake 1.15 required for boost 1.71
```
ajelinski, line 3: Remove this comment; it described why boost 1.71 couldn't have been used.

pnr/nextpnr/ecp5/meta.yaml
```diff
@@ -1,7 +1,9 @@
 # Use `conda-build-prepare` before building for a better version string.
 {% set version = '%s_%04i_%s'|format(GIT_DESCRIBE_TAG|replace('v', '') or '0.X.X', GIT_DESCRIBE_NUMBER|int, GIT_DESCRIBE_HASH or 'gUNKNOWN') %}
 # cmake 1.15 required for boost 1.71
```
ajelinski, line 3: Remove this one too.

pnr/nextpnr/ecp5/meta.yaml
```diff
@@ -26,31 +28,31 @@ requirements:
...
     - boost     {{ boost_version }} [not win]
     - boost-cpp {{ boost_version }} [not win]
     - libboost  {{ boost_version }} [not win]
     - py-boost  {{ boost_version }} [not win]
+    - python {{ python_version }}
   run:
     - libgcc-ng     [linux]
     - libstdcxx-ng  [linux]
     - prjtrellis
     - {{ pin_compatible('python', min_pin='x.x', max_pin='x.x') }}
```
ajelinski, line 52: Remove this dependency. It adds a Python requirement depending on which is used in `host` Python but it's redundant when used along `python {{ python_version }}`.

ajelinski: Also remove the same requirement in nextpnr-generic and nextpnr-ice40.

pnr/nextpnr/generic/meta.yaml
```diff
@@ -1,7 +1,9 @@
 # Use `conda-build-prepare` before building for a better version string.
 {% set version = '%s_%04i_%s'|format(GIT_DESCRIBE_TAG|replace('v', '') or '0.X.X', GIT_DESCRIBE_NUMBER|int, GIT_DESCRIBE_HASH or 'gUNKNOWN') %}
 # cmake 1.15 required for boost 1.71
```
ajelinski, line 3: Remove

ajelinski (changes requested):
I've missed a few more issues.

bit/prjtrellis/meta.yaml
mtdudek: I've fixed it

ajelinski (approved):

ajelinski:
Without `pin_compatible` in `run` requirements there are conflicts in `build` requirements because:
* Conda chooses newer fontconfig which requires newer libxml2 than the latest one on LiteX-Hub so the one from main channel is used,
* libxml2 from the main channel causes a conflict.
Both of these are highly unreasonable but it is how it is.

Since having `pin_compatible` in run have the same result as `python {{ python_version }}` we might want to replace these. The CI with such a change has already started: https://github.com/hdl/conda-eda/actions/runs/1056182265

ajelinski:
`nextpnr-ecp5` was expected to fail since it couldn't use new `prjtrellis` package with the latest Boost (no upload in the cross-repository PRs). Both of these now use Boost 1.73 while the old `prjtrellis` package requires Boost 1.67 to be used.

I'm merging this PR since it worked in a CI run from a branch where it could use the modified `prjtrellis`: https://github.com/hdl/conda-eda/runs/3134196881?check_suite_focus=true

PR #119, opened 2021-07-21
Yosys plugins symbiflow

mtdudek:
Adds package with symbiflow-plugins for yosys.
There is package very similar to this one, but that one uses sybmiflow-yosys and not main stream one.

### DIFF
--- a/.github/workflows/build-packages.yml
+++ b/.github/workflows/build-packages.yml
@@ -18,7 +18,7 @@ on:
 env:
   ANACONDA_TOKEN: ${{ secrets.ANACONDA_TOKEN }}
   ANACONDA_USER: ${{ secrets.ANACONDA_USER }}
-  NUM_OF_JOBS: 70
+  NUM_OF_JOBS: 71
 defaults:
   run:
     shell: bash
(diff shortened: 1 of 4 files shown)


.github/workflows/build-packages.yml
```diff
@@ -421,7 +434,7 @@ jobs:
     env:
       PACKAGE: "xilinx/vivado"
       OS_NAME: "linux"
-      VERSIONS: ${{ format('{0}{1}', github.workspace, '/xilinx/vivado/versions') }} 
+      VERSIONS: ${{ format('{0}{1}', github.workspace, '/xilinx/vivado/versions') }}
```
ajelinski, line 437: Undo this change.

.github/workflows/build-packages.yml
```diff
@@ -676,7 +689,7 @@ jobs:
   #47
   nextpnr-ice40-windows:
     runs-on: "windows-latest"
-    needs: ["icestorm-windows"] 
+    needs: ["icestorm-windows"]
```
ajelinski, line 692: Undo this change.

ajelinski (commented):
Try not to modify anything else than necessary to make the change stated by the commit. Add a new commit that fixes them explicitly if you see anything that should be changed.

mtdudek:
Removed unnecessary changes

ajelinski (approved):

Issue #120, opened 2021-07-23
nextpnr-ice40/generic: Building on Windows fails due to `undeclared identified`


ajelinski:
The error building both `nextpnr-ice40` and `nextpnr-generic` is the same and comes from `common/bits.h`:
```
  c:\users\runner\conda-env\conda-bld\nextpnr-generic_1627032814927\work\common\bits.h(54): error C2065: 'v': undeclared identifier [%SRC_DIR%\nextpnr-generic.vcxproj]
  c:\users\runner\conda-env\conda-bld\nextpnr-generic_1627032814927\work\common\bits.h(66): error C2065: 'v': undeclared identifier [%SRC_DIR%\nextpnr-generic.vcxproj]
```

PR #121, opened 2021-07-23
nextpnr: Fix macOS issues and partly Windows

ajelinski:
macOS issues are fully fixed by raising compiler restriction that was once required to build these packages.

After solving one problem on Windows another came up so the specific issue was created for it: https://github.com/hdl/conda-eda/issues/120 .


ajelinski:
The PR fixes macOS builds and clarifies Windows issues so I'll merge it.

PR #124, opened 2021-08-10
pnr: Add nextpnr-nexus

MateuszKosmala:
PR adds nextpnr-nexus to Place and Route.

### DIFF
--- a/.github/workflows/build-packages.yml
+++ b/.github/workflows/build-packages.yml
@@ -18,7 +18,7 @@ on:
 env:
   ANACONDA_TOKEN: ${{ secrets.ANACONDA_TOKEN }}
   ANACONDA_USER: ${{ secrets.ANACONDA_USER }}
-  NUM_OF_JOBS: 71
+  NUM_OF_JOBS: 72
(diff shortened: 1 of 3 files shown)


ajelinski (approved):
It looks good. The recipe could be cleaned from unnecessary parts (e.g. Windows deps) but let's leave it because some day it might be needed on Windows too.

ajelinski:
BTW. The package is going to be built with Python 3.7, I guess you don't need it to work with Python 3.8?

MateuszKosmala:
Yeah, That version suits me.

ajelinski:
The job failed. You might check if adding `-DCMAKE_INSTALL_PREFIX=/` helps as it's used in other `nextpnr`s: https://github.com/hdl/conda-eda/blob/master/pnr/nextpnr/ice40/build.sh#L6

MateuszKosmala:
Thanks, I changed that.

ajelinski:
I'll handle failing `symbiflow-vtr-gui-linux` in a separate PR.

PR #125, opened 2021-08-11
symbiflow-vtr-gui: Stick to GCC <11

ajelinski:
Tool's build scripts haven't been adapted to the latest GCC yet.

Co-authored by: Maciej Dudek <mdudek@antmicro.com>


ajelinski:
The job has been making the whole CI fail since 5th of August.
GCC 11.1 was used in all of these failing builds and the latest successful one used GCC 9.

In the CI GCC 10.3.0 was used successfully so the problem seems to be strictly related to the GCC 11.

PR #126, opened 2021-08-12
syn: yosys-plugins: depend on yosys and not symbiflow-yosys

acomodi:


acomodi:
The master run failed for a connection issue apparently, but all the builds were successful

ajelinski (approved):

ajelinski:
The only error is `urllib.error.HTTPError: HTTP Error 403: rate limit exceeded` in `master-package`.
I've seen it also in a CI run from `master` a few days ago so it's an unrelated issue.

PR #131, opened 2021-09-16
yosys-plugins-symbiflow; symbiflow-yosys-plugins: update patch

kamilrakoczy:
This PR updates yosys-plugins-symbiflow patch to match changes in current master revision.


ajelinski (approved):
LGTM

PR #132, opened 2021-09-20
yosys-uhdm: fix missing libffi in host

kamilrakoczy:
This PR fixes: ``libffi.so.7: cannot open shared object file: No such file or directory``: https://github.com/chipsalliance/sv-tests/issues/1758


ajelinski (approved):
LGTM; using `libffi` installed in the OS during building is clearly a problem here since the latest Conda package only provides `libffi.so.6` while `yosys-uhdm` was linked against `libffi.so.7`.

PR #134, opened 2021-09-23
Update jobs missed to use Ubuntu 20.04

PiotrZierhoffer:
This is to verify prjoxide build


PiotrZierhoffer:
Projxide works, merging

PR #140, opened 2021-10-01
nextpnr: Fix package versions after switching to the upstream repository

### LINKED ISSUE #139: nextpnr-nexus: default "main" not tracking most recent build

ajelinski:
The upstream repository doesn't have any tags so `conda-build-prepare`
tags the initial commit with the `v0.0` tag. The `timvideos/nextpnr`
fork has the initial commit tagged differently which is what causes the
version naming change. It results in the old packages being preferred by
Conda.

Let's use the `extra.tags` file to tag the initial commit similarly as
in the `timvideos` fork during the repository preparation.

Fixes #139.


PiotrZierhoffer:
Lgtm, merge on green!

PiotrZierhoffer (approved):

PR #143, opened 2021-10-04
Move zachjs sv2v from conda-misc

PiotrZierhoffer:


PiotrZierhoffer:
Zachjs-sv2v passed and it's not a dependency for anything. Merging.

PR #48, opened 2020-11-24
WIP: add symbiflow-toolchain package

acomodi:
Signed-off-by: Alessandro Comodi <acomodi@antmicro.com>

This PR adds the symbiflow-toolchain binaries from the packages generated by symbiflow-arch-defs.
Currently, a hardcoded link is provided for the binaries, hence the `WIP`.

PR #145, opened 2021-11-30
toolchain: add symbiflow-toolchain package

acomodi:

This supersedes https://github.com/hdl/conda-eda/pull/48, and adds the symbiflow toolchain package


toolchain/symbiflow-toolchain/build.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+mkdir -p $PREFIX
+
+curl -s https://storage.googleapis.com/symbiflow-arch-defs-gha/symbiflow-toolchain-latest | xargs wget -qO- | tar -xJC $PREFIX
```
mithro, line 5: We **really** need to make sure that there is a connection between these assets and the symbiflow-arch-defs git revision and not use the random latest version....

toolchain/symbiflow-toolchain/meta.yaml
```diff
@@ -0,0 +1,30 @@
+{% set version = '%s_%04i_%s'|format(GIT_DESCRIBE_TAG|replace('v','') or '0.X', GIT_DESCRIBE_NUMBER|int, GIT_DESCRIBE_HASH or 'gUNKNOWN') %}
+
+package:
+  name: symbiflow-toolchain
+  version: {{ version }}
+
+source:
+  git_url: https://github.com/verilog-to-routing/vtr-verilog-to-routing.git
```
mithro, line 8: This should be `https://github.com/SymbiFlow/symbiflow-arch-defs.git`

PiotrZierhoffer:
As we've been discussing this internally, our current approach is to:

- have a separate CI flow that checks the latest revision and rebuilds the package with the same revision name
- have the package listed explicitly in target project's  environment

This PR is a stub we begin with, based on the work we started ~1 year ago

mithro:
Is there a simple way to go from symbiflow-arch-defs git revision number to the build CI output?

If so, you could use that process so that the contents are determined by the symbiflow-arch-defs git revision the package is built with?

acomodi:
Actually the git revision number is present within the link contained in `https://storage.googleapis.com/symbiflow-arch-defs-gha/symbiflow-toolchain-latest`.

For instance, the current latest link is the following:
```
https://www.googleapis.com/download/storage/v1/b/symbiflow-arch-defs/o/artifacts%2Fprod%2Ffoss-fpga-tools%2Fsymbiflow-arch-defs%2Fcontinuous%2Finstall%2F474%2F20211130-000503%2Fsymbiflow-arch-defs-install-8842ce79.tar.xz?generation=1638277014446946&alt=media
```

Where `8842ce79` is the sha of the most recent commit that generated the arch files in arch-defs: https://github.com/SymbiFlow/symbiflow-arch-defs/commit/8842ce790f1fa32b975ed7564ceb908141511acc.

The above link also contains the build date, so we may probably extract that as well from there.

toolchain/symbiflow-toolchain/meta.yaml
acomodi: Fixed

toolchain/symbiflow-toolchain/build.sh
mithro: So this needs to download the results which match the git revision used to generate the package metadata. That info should be available in the shell environment somewhere...

mithro:
> Actually the git revision number is present within the link contained in `https://storage.googleapis.com/symbiflow-arch-defs-gha/symbiflow-toolchain-latest`.
> 
> For instance, the current latest link is the following:
> 
> ```
> https://www.googleapis.com/download/storage/v1/b/symbiflow-arch-defs/o/artifacts%2Fprod%2Ffoss-fpga-tools%2Fsymbiflow-arch-defs%2Fcontinuous%2Finstall%2F474%2F20211130-000503%2Fsymbiflow-arch-defs-install-8842ce79.tar.xz?generation=1638277014446946&alt=media
> ```
> 
> Where `8842ce79` is the sha of the most recent commit that generated the arch files in arch-defs: [SymbiFlow/symbiflow-arch-defs@8842ce7](https://github.com/SymbiFlow/symbiflow-arch-defs/commit/8842ce790f1fa32b975ed7564ceb908141511acc).
> 
> The above link also contains the build date, so we may probably extract that as well from there.

That is the wrong way around. You want to go from git revision ID to a build result (not from a build result to a git revision ID).

PR #148, opened 2021-12-10
pnr: arachne: remove symbiflow-yosys from deps

acomodi:

The symbiflow-yosys package will soon be deprecated. Moreover, the pnr tool should not depend on the synthesis tool, so it should be completely removed


PiotrZierhoffer:
Let's not merge this - Arachne is dropped completely.

Issue #160, opened 2022-02-01
add open_pdks package


proppy:
upstream: http://opencircuitdesign.com/open_pdks/
minimal recipe per http://opencircuitdesign.com/open_pdks/install.html:
```
./configure --enable-sky130-pdk=/path/to/skywater-pdk
make
make install
```
depends: `skywater-pdk` #159 https://anaconda.org/LiteX-Hub/magic

proppy:
@ajelinski anything else we need to do to get it uploaded?

ajelinski:
It'll be uploaded automatically by the `master-package` job once all the other jobs succeed: https://github.com/hdl/conda-eda/actions/runs/1845871121

proppy:
This does seems related to the `UNAMED` error we say in #161, see:
https://github.com/RTimothyEdwards/magic/blob/bfdf1227a704dbbfb78f42f69965ccc2c81989c1/commands/CmdE.c#L1051
https://github.com/RTimothyEdwards/magic/blob/bfdf1227a704dbbfb78f42f69965ccc2c81989c1/commands/CmdE.c#L1017

ajelinski:
Also this might be helpful:
```
2022-02-15T09:30:40.4701151Z Traceback (most recent call last):
2022-02-15T09:30:40.4701428Z   File "../common/foundry_install.py", line 2245, in <module>
2022-02-15T09:30:40.4701686Z     cwd = destlibdir,
2022-02-15T09:30:40.4701970Z   File "../common/foundry_install.py", line 277, in subprocess_run
2022-02-15T09:30:40.4702263Z     raise SystemError("".join(emsg))
2022-02-15T09:30:40.4702895Z SystemError: Command magic failed with exit code: -11
2022-02-15T09:30:40.4703391Z   magic -dnull -noconsole $SRC_DIR/sky130/sky130A/libs.ref/sky130_fd_sc_hd/spice/generate_magic.tcl
```

BTW There are a lot of these `Cell "sky130_fd_sc_hd__fill_4" is used but not defined in this file.` errors even in the successful job so they might not be really relevant. I guess this " Rewinding input.  Cells may have been instanced before they were defined." must be true but then I'm not sure why this isn't just a warning.

Is it a problem observable outside this repository's CI?

proppy:
I suspect the actual fatal error is:
```
  Please name the cell before extracting.
```
[...]

PR #161, opened 2022-02-01
misc: add skywater-pdk

### LINKED ISSUE #159: add skywater-pdk package
upstream: https://github.com/google/skywater-pdk
minimal recipe http://opencircuitdesign.com/open_pdks/install.html:
```
git clone https://github.com/google/skywater-pdk
cd skywater-pdk
git submodule init libraries/sky130_fd_io/latest
git submodule init libraries/sky130_fd_pr/latest
git submodule init libraries/sky130_fd_sc_hd/latest
git submodule init libraries/sky130_fd_sc_hvl/latest
git submodule update
make timing
```

proppy:
Fixes #159 

- add recipe for `skywater-pdk` python scripts
- add recipes for python transitives dependencies: `dataclasses-json` `typing-inspect`
- add subpackages recipes standard cells libraries recommended by http://opencircuitdesign.com/open_pdks/install.html:  `sky130_fd_sc_hd` `sky130_fd_sc_hvl` `sky130_fd_io` and `sky130_fd_pr`
- generate timing for `sky130_fd_sc_hd` `sky130_fd_sc_hvl` 
- install in `$PREFIX/share/skywater-pdk`

/cc @olofk @mithro

### DIFF
--- a/misc/skywater-pdk/build-lib.sh
+++ b/misc/skywater-pdk/build-lib.sh
@@ -1,0 +1,21 @@
+#!/bin/bash
+set -ex
+
+LIB=${PKG_NAME#skywater-pdk.}
+LIB_DIR=libraries/$LIB/v$PKG_VERSION
+mkdir -p $PREFIX/share/skywater-pdk/$LIB_DIR/
+(cd $PREFIX/share/skywater-pdk/libraries/$LIB && ln -s v$PKG_VERSION latest)
+cp $SRC_DIR/$LIB_DIR/LICENSE $PREFIX/share/skywater-pdk/$LIB_DIR/
+cp -r $SRC_DIR/$LIB_DIR/cells $PREFIX/share/skywater-pdk/$LIB_DIR/
+if [[ -d $SRC_DIR/$LIB_DIR/models ]]; then
+  cp -r $SRC_DIR/$LIB_DIR/models $PREFIX/share/skywater-pdk/$LIB_DIR/
+fi
+if [[ -d $SRC_DIR/$LIB_DIR/tech ]]; then
+  cp -r $SRC_DIR/$LIB_DIR/tech $PREFIX/share/skywater-pdk/$LIB_DIR/
+fi
+if [[ -d $SRC_DIR/$LIB_DIR/timing ]]; then
+  $PYTHON -m skywater_pdk.liberty $LIB_DIR
+  $PYTHON -m skywater_pdk.liberty $LIB_DIR all
+  $PYTHON -m skywater_pdk.liberty $LIB_DIR all --ccsnoise
+  cp -r $SRC_DIR/$LIB_DIR/timing $PREFIX/share/skywater-pdk/$LIB_DIR/
+fi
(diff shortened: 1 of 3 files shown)


mithro (commented):
The current way to "install" the skywater-pdk is using the open_pdks script?

proppy:
@mithro I was initially thinking that the `open_pdks` conda package could stack on top see #160.

proppy:
But that's true we could ship all pdks with the `open_pdks` package rather than introducing direct dependencies; it would make the recipe for open_pdks "bigger" as we add more pdks (asap7, etc).

proppy:
I think the "real test" is to know if PDKs (skywater and others) could have reverse downstream (`run` or `build`) dependencies that are not `open_pdks`, i.e: are there tools that consume PDKs with their original layout.

PiotrZierhoffer:
@proppy I merged some fixes to the current CI, so if you rebase onto current master you should have a green pass of tests.

proppy:
@PiotrZierhoffer sure, I have other improvement I want to add.

- make the package `noarch`
- split the python stuff into a separate `build` dep

proppy:
@PiotrZierhoffer PTAL rebased

proppy:
We should probably use a common prefix for all the subpackages.

proppy:
And split test into each subpackages.

misc/typing-inspect/meta.yaml
```diff
@@ -0,0 +1,41 @@
+{% set version = "0.7.1" %}
```
proppy, line 1: we don't actually need this, since it's available on conda-forge:
https://anaconda.org/conda-forge/typing_inspect

misc/dataclasses-json/meta.yaml
```diff
@@ -0,0 +1,45 @@
+{% set name = "dataclasses-json" %}
```
proppy, line 1: we don't actually need this, since it's available on conda-forge:
https://anaconda.org/conda-forge/dataclasses-json

misc/skywater-pdk/build-lib.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+set -ex
+
+LIB=${PKG_NAME#skywater-pdk.}
+LIB_DIR=libraries/$LIB/v$PKG_VERSION
+mkdir -p $PREFIX/share/skywater-pdk/$LIB_DIR/
```
proppy, line 6: we also need a `latest` symlink for `open_pdks` to be happy, see https://github.com/RTimothyEdwards/open_pdks/blob/dd84b77085a5bffba8d5493e82137322388997b9/sky130/Makefile.in#L918

misc/typing-inspect/meta.yaml
proppy: Done

misc/dataclasses-json/meta.yaml
proppy: Done

misc/skywater-pdk/build-lib.sh
proppy: Done

proppy:
dropping in favor of consolidating in #164

PR #162, opened 2022-02-02
moore: Update building script with LLVM

PiotrZierhoffer:
Moore added a requirement of manually building LLVM.

It might be a hassle, but let's give it a try


PiotrZierhoffer:
This is definitely not enough. Will investigate this further, but for now we'll skip

PR #165, opened 2022-02-03
pnr: add openroad

### LINKED ISSUE #150: add openroad package
It would be neat to have [openroad](https://github.com/The-OpenROAD-Project/OpenROAD) available as a conda package.

Maybe doing something similar to http://yowasp.org/ for better cross-platform support?

proppy:
Fixes #150 

- distribute `openroad.app` `openroad.sta` and openroad.libsta` as subpackages
- distribute `openroad` meta package with `.app` and `.sta`
- patch to remove testing (failed to link against `rt`)
- user conda-eda convention (build from git/master, use `conda-build-prepare` for version string)

@PiotrZierhoffer @mithro for review

@maliberty fyi

### DIFF
--- a/pnr/openroad/build.sh
+++ b/pnr/openroad/build.sh
@@ -1,0 +1,24 @@
+#!/bin/bash
+
+
+# Copyright 2022 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -ex
+
+mkdir build
+cd build
+cmake -DCMAKE_FIND_ROOT_PATH="$BUILD_PREFIX;$PREFIX" -DCMAKE_FIND_ROOT_PATH_MODE_INCLUDE=ONLY -DCMAKE_FIND_ROOT_PATH_MODE_LIBRARY=ONLY -DCMAKE_FIND_ROOT_PATH_MODE_PROGRAM=ONLY -DUSE_SYSTEM_BOOST=ON -DINSTALL_LIBOPENSTA=OFF -DBUILD_GUI=OFF -DCMAKE_INSTALL_PREFIX=$PREFIX  ..
+make -j$CPU_COUNT
+make -j$CPU_COUNT install
--- a/pnr/openroad/meta.yaml
+++ b/pnr/openroad/meta.yaml
@@ -1,0 +1,65 @@
+# Use `conda-build-prepare` before building for a better version string.
+{% set version = '%s_%04i_%s'|format(GIT_DESCRIBE_TAG|replace('v', '')|replace('-', '')|default('0.X'), GIT_DESCRIBE_NUMBER|int, GIT_DESCRIBE_HASH|default('gUNKNOWN')) %}
+
+package:
+  name: openroad
+  version: {{ version }}
+
+source:
+  - git_url: https://github.com/The-OpenROAD-Project/OpenROAD.git
+    git_rev: master
+    patches:
+      - odb-drt-disable-tests.patch
+      - sta-add-install-options.patch
+      - gui-find-qt-only-if-enabled.patch
+
+build:
+  # number: 202202031935
+  number: {{ environ.get('DATE_NUM') }}
+  # string: 20220203_1935
+  string: {{ environ.get('DATE_STR') }}
+  script_env:
+    - CI
+  ignore_run_exports_from:
+    # header-only libraries
+    - lemon
+    - fmt
+
+requirements:
+  build:
+    - make
+    - cmake >=3.19
+    - swig 4.0.1
+    - bison
+    - doxygen
+    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
+  host:
+    - python
+    - boost-cpp 1.76
+    - eigen 3.3
+    - lemon 1.3.1
+    - spdlog 1.8.1
+    - readline
+    - tk
+    - zlib
+    - fmt
+    - libgomp
+  run:
+    # missing run_exports
+    - {{ pin_compatible('boost-cpp', min_pin='x.x', max_pin='x.x') }}
+    - {{ pin_compatible('python', min_pin='x.x', max_pin='x.x') }}    
+    
+test:
+  commands:
+    - openroad -version
+    - sta -version
+
+about:
+  home: https://theopenroadproject.org/
+  license: BSD-3-Clause
+  summary: OpenROAD is an unified application implementing an RTL-to-GDS Flow.
+
+extra:
+  recipe-maintainers:
+    - proppy
--- a/.github/workflows/build-packages.yml
+++ b/.github/workflows/build-packages.yml
@@ -18,7 +18,7 @@ on:
 env:
   ANACONDA_TOKEN: ${{ secrets.ANACONDA_TOKEN }}
   ANACONDA_USER: ${{ secrets.ANACONDA_USER }}
-  NUM_OF_JOBS: 68
+  NUM_OF_JOBS: 69
 defaults:
   run:
     shell: bash
@@ -763,6 +763,16 @@ jobs:
     runs-on: "ubuntu-20.04"
     env:
       PACKAGE: "misc/open_pdks"
+      OS_NAME: "linux"
+    steps:
+      - uses: actions/checkout@v2
+      - uses: hdl/conda-ci@master
+
+  #69
+  openroad-linux:
+    runs-on: "ubuntu-20.04"
+    env:
+      PACKAGE: "pnr/openroad"
       OS_NAME: "linux"
     steps:
       - uses: actions/checkout@v2
--- a/pnr/openroad/condarc
+++ b/pnr/openroad/condarc
@@ -1,0 +1,2 @@
+channels:
+  - conda-forge
--- a/pnr/openroad/gui-find-qt-only-if-enabled.patch
+++ b/pnr/openroad/gui-find-qt-only-if-enabled.patch
@@ -1,0 +1,23 @@
+diff --git a/src/gui/CMakeLists.txt b/src/gui/CMakeLists.txt
+index 204e14956..9e8a1bac3 100755
+--- a/src/gui/CMakeLists.txt
++++ b/src/gui/CMakeLists.txt
+@@ -1,15 +1,13 @@
+ # Allow users to not build without the gui even if Qt is found
+ option(BUILD_GUI "Build the GUI" ON)
+ 
+-# If Qt is not installed there will not be cmake
+-# support for the package so this needs to be "quiet".
+-find_package(Qt5 QUIET COMPONENTS Core Widgets)
+-
+ include("openroad")
+ set(CMAKE_INCLUDE_CURRENT_DIR ON)
+ 
+-if (Qt5_FOUND AND BUILD_GUI)
++if (BUILD_GUI)
+   message(STATUS "GUI is enabled")
++  find_package(Qt5 QUIET COMPONENTS Core Widgets)
++
+   set(CMAKE_AUTOMOC ON)
+   set(CMAKE_AUTORCC ON)
+   set(CMAKE_AUTOUIC ON)
--- a/pnr/openroad/odb-drt-disable-tests.patch
+++ b/pnr/openroad/odb-drt-disable-tests.patch
@@ -1,0 +1,60 @@
+diff --git a/src/drt/CMakeLists.txt b/src/drt/CMakeLists.txt
+index ce93a063e..2c3afc754 100644
+--- a/src/drt/CMakeLists.txt
++++ b/src/drt/CMakeLists.txt
+@@ -159,43 +159,6 @@ target_compile_options(drt
+     $<$<CXX_COMPILER_ID:Clang>:-Wall -pedantic -Wcast-qual -Wredundant-decls -Wformat-security -Wno-gnu-zero-variadic-macro-arguments>
+   )
+ 
+-############################################################
+-# Unit testing
+-############################################################
+-enable_testing()
+-
+-add_executable(trTest
+-  ${FLEXROUTE_HOME}/test/gcTest.cpp
+-  ${FLEXROUTE_HOME}/test/fixture.cpp
+-)
+-
+-target_include_directories(trTest
+-  PRIVATE
+-  ${FLEXROUTE_HOME}/src
+-  ${OPENROAD_HOME}/include
+-)
+-
+-target_link_libraries(trTest
+-  drt
+-  odb
+-)
+-
+-# Use the shared library if found.  We need to pass this info to
+-# the code to select the corresponding include.  Using the shared
+-# library speeds up compilation.
+-if (Boost_unit_test_framework_FOUND)
+-  message(STATUS "Boost unit_test_framework library found")
+-  target_link_libraries(trTest
+-    Boost::unit_test_framework
+-  )
+-  target_compile_definitions(trTest
+-    PRIVATE
+-    HAS_BOOST_UNIT_TEST_LIBRARY
+-  )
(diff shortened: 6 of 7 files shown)


pnr/openroad/meta.yaml
```diff
@@ -0,0 +1,77 @@
...
+  name: openroad
+  version: {{ version }}
+
+source:
+  - git_url: https://github.com/The-OpenROAD-Project/OpenROAD.git
+    git_rev: master
+    patches:
+      - disable-tests.patch
+  - git_url: https://github.com/The-OpenROAD-Project/OpenSTA.git
+    git_rev: master
```
proppy, line 14: is that ok to pull master rather than using the version pinned as a submodules?

pnr/openroad/meta.yaml
```diff
@@ -0,0 +1,77 @@
...
+    - {{ compiler('c') }}    
+    - {{ compiler('cxx') }}
+  host:
+    - boost
+    - eigen
+    - lemon
+    - spdlog
+    - readline
+  run:
+    openroad.app
```
proppy, line 41: that should be a list

pnr/openroad/meta.yaml
proppy: we don't actually need to do this, since `git_url` already does a recursive clone https://github.com/conda/conda-build/blob/master/conda_build/source.py#L328

pnr/openroad/meta.yaml
```diff
@@ -0,0 +1,77 @@
...
+source:
+  - git_url: https://github.com/The-OpenROAD-Project/OpenROAD.git
+    git_rev: master
+    patches:
+      - disable-tests.patch
+  - git_url: https://github.com/The-OpenROAD-Project/OpenSTA.git
+    git_rev: master
+    folder: src/sta
+  - git_url: https://github.com/The-OpenROAD-Project/abc.git
+    git_rev: master
```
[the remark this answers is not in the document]
proppy: we don't actually need to do this, since `git_url` already does a recursive clone https://github.com/conda/conda-build/blob/master/conda_build/source.py#L328

pnr/openroad/meta.yaml
proppy: Done

pnr/openroad/meta.yaml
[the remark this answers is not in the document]
proppy: Done

pnr/openroad/meta.yaml
proppy: Done

proppy:
@PiotrZierhoffer does the ci (and `conda-build-prepare`) handle subpackages well? (This does build for me locally without it)

proppy:
@PiotrZierhoffer @mithro PTAL, got rid of the subpackages, and CI now passes: https://github.com/hdl/conda-eda/runs/5075946231?check_suite_focus=true

pnr/openroad/build.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
proppy: Done

pnr/openroad/meta.yaml
```diff
@@ -0,0 +1,53 @@
...
+    - readline
+
+test:
+  commands:
+    - openroad -version
+    - sta -version
+
+about:
+  home: https://theopenroadproject.org/
+  license: BSD-3-Clause
```
proppy, line 49: `summary` missing

pnr/openroad/build.sh
```diff
@@ -0,0 +1,23 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -ex
+
+mkdir build
+cd build
+cmake -DUSE_SYSTEM_BOOST=ON -DALLOW_WARNINGS=ON -DINSTALL_LIBOPENSTA=OFF  -DCMAKE_INSTALL_PREFIX=$PREFIX ..
```
maliberty, line 21: -DALLOW_WARNINGS=ON is now the default

pnr/openroad/build.sh
```diff
@@ -0,0 +1,23 @@
...
+make -j$CPU_COUNT
+make install
```
maliberty, line 23: make -j$CPU_COUNT install

pnr/openroad/meta.yaml
```diff
@@ -0,0 +1,53 @@
...
+
+package:
+  name: openroad
+  version: {{ version }}
+
+source:
+  - git_url: https://github.com/The-OpenROAD-Project/OpenROAD.git
+    git_rev: master
+    patches:
+      - odb-drt-disable-tests.patch
```
maliberty, line 12: What are these?

proppy: Sorry for the confusion, I also plan to submit those patches upstream (after reworking them a little bit, for example: this particular one would need to be put behind a additional cmake flag).

proppy:
@ajelinski PTAL: rebased, bumped `NUM_OF_JOBS` and cleaned up the history.

pnr/openroad/meta.yaml
proppy: Done

pnr/openroad/build.sh
proppy: Removed.

pnr/openroad/meta.yaml
proppy: Ack, will submit a PR upstream once I rework the patch.

pnr/openroad/build.sh
proppy: Removed.

pnr/openroad/meta.yaml
```diff
@@ -0,0 +1,59 @@
...
+    - {{ compiler('cxx') }}
+  host:
+    - python
+    - boost
+    - eigen
+    - lemon
+    - spdlog
+    - readline
+  run:
+    - python
```
ajelinski, line 40: Can the tool be run with any Python version after it's built? If not, it will be necessary to restrict its version and also perhaps add a few jobs for each Python version which is going to be supported as it's done with the `syn/yosys` package.

ajelinski (changes requested):
There's a problem with Python headers during building:
```
2022-02-15T13:38:17.3751609Z [ 84%] Building CXX object src/odb/src/swig/python/CMakeFiles/odbpy.dir/CMakeFiles/odbpy.dir/odbPYTHON_wrap.cxx.o
2022-02-15T13:38:17.3754823Z cd $SRC_DIR/build/src/odb/src/swig/python && $BUILD_PREFIX/bin/x86_64-conda-linux-gnu-c++ -DBOOST_ALL_NO_LIB -DFMT_SHARED -DSPDLOG_COMPILED_LIB -DSPDLOG_SHARED_LIB -I$SRC_DIR/include -I$SRC_DIR/src/odb/include/odb -I$SRC_DIR/src/odb/include -I$SRC_DIR/src/utl/include -I$SRC_DIR/src/odb/src/def/def -I$SRC_DIR/src/odb/src/def/defzlib -I$SRC_DIR/src/odb/src/lef/lef -I$SRC_DIR/src/odb/src/lef/lefzlib -I$SRC_DIR/src/odb/src/swig/common/. -isystem /usr/include/python3.8 -fvisibility-inlines-hidden -fmessage-length=0 -march=nocona -mtune=haswell -ftree-vectorize -fPIC -fstack-protector-strong -fno-plt -O2 -ffunction-sections -pipe -isystem $PREFIX/include -fdebug-prefix-map=$SRC_DIR=/usr/local/src/conda/openroad-2.0_2977_ga7ea081ae -fdebug-prefix-map=$PREFIX=/usr/local/src/conda-prefix -O3 -DNDEBUG -Wall -Wredundant-decls -Wformat-security -Wno-sign-compare -std=c++17 -Wno-cast-qual -Wno-missing-braces -MD -MT src/odb/src/swig/python/CMakeFiles/odbpy.dir/CMakeFiles/odbpy.dir/odbPYTHON_wrap.cxx.o -MF CMakeFiles/odbpy.dir/CMakeFiles/odbpy.dir/odbPYTHON_wrap.cxx.o.d -o CMakeFiles/odbpy.dir/CMakeFiles/odbpy.dir/odbPYTHON_wrap.cxx.o -c $SRC_DIR/build/src/odb/src/swig/python/CMakeFiles/odbpy.dir/odbPYTHON_wrap.cxx
2022-02-15T13:38:17.3870357Z In file included from /usr/include/python3.8/Python.h:8,
2022-02-15T13:38:17.3871142Z                  from /home/runner/work/conda-eda/conda-eda/workdir/conda-env/conda-bld/openroad_1644931048506/work/build/src/odb/src/swig/python/CMakeFiles/odbpy.dir/odbPYTHON_wrap.cxx:178:
2022-02-15T13:38:17.3871800Z /usr/include/python3.8/pyconfig.h:3:12: fatal error: x86_64-linux-gnu/python3.8/pyconfig.h: No such file or directory
2022-02-15T13:38:17.3873722Z     3 | #  include <x86_64-linux-gnu/python3.8/pyconfig.h>
2022-02-15T13:38:17.3874063Z       |            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
2022-02-15T13:38:17.3874682Z compilation terminated.
2022-02-15T13:38:17.3888740Z make[2]: *** [src/odb/src/swig/python/CMakeFiles/odbpy.dir/build.make:92: src/odb/src/swig/python/CMakeFiles/odbpy.dir/CMakeFiles/odbpy.dir/odbPYTHON_wrap.cxx.o] Error 1
2022-02-15T13:38:17.3889231Z make[1]: *** [CMakeFiles/Makefile2:1853: src/odb/src/swig/python/CMakeFiles/odbpy.dir/all] Error 2
2022-02-15T13:38:17.3889546Z make[1]: *** Waiting for unfinished jobs....
2022-02-15T13:38:19.3757027Z make: *** [Makefile:139: all] Error 2
```
It seems to be caused by using Python installed in the OS which is also true for Swig: https://github.com/hdl/conda-eda/runs/5199459111?check_suite_focus=true#step:3:2001

You might experiment with making swig a `host` dependency and for Python maybe `python 3.8` will help as I see now 3.7 is installed as a dependency: https://github.com/hdl/conda-eda/runs/5199459111?check_suite_focus=true#step:3:1727

pnr/openroad/meta.yaml
proppy: actually `python` shouldn't be required as a `run` dependency as the python interpreter is embedded.

proppy:
> There's a problem with Python headers during building:

I'd assume that's because we're missing `python-dev` as a dep?

pnr/openroad/meta.yaml
```diff
@@ -0,0 +1,59 @@
...
+  # number: 202202031935
+  number: {{ environ.get('DATE_NUM') }}
+  # string: 20220203_1935
+  string: {{ environ.get('DATE_STR') }}
+  script_env:
+    - CI
+
+requirements:
+  build:
+    - cmake
```
proppy, line 26: we should add >= 3.14:
https://github.com/The-OpenROAD-Project/OpenROAD/blob/master/CMakeLists.txt#L23

pnr/openroad/meta.yaml
```diff
@@ -0,0 +1,59 @@
...
+    - swig
```
proppy, line 27: seems like >= 3.0 is required:
https://github.com/The-OpenROAD-Project/OpenSTA/blob/8627c6826d752ed864faafb71461d3c24ae66655/CMakeLists.txt#L293

pnr/openroad/meta.yaml
```diff
@@ -0,0 +1,59 @@
...
+    - bison
+    - doxygen
+    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
+  host:
```
proppy, line 32: @maliberty is the package functional without `cudd`? https://github.com/The-OpenROAD-Project/OpenSTA/blob/8627c6826d752ed864faafb71461d3c24ae66655/CMakeLists.txt#L394

proppy, line 32: we should use the same version used here: https://github.com/The-OpenROAD-Project/OpenROAD/blob/af1f38c9f23493ae1f604566066e2b9d24e336b2/etc/DependencyInstaller.sh#L5

pnr/openroad/meta.yaml
proppy: ah, but we probably still need it for being able to resolve libpython at runtime.

proppy: trying to pin the version and see if it helps.

proppy:
> It seems to be caused by using Python installed in the OS which is also true for Swig: https://github.com/hdl/conda-eda/runs/5199459111?check_suite_focus=true#step:3:2001

Reading thru: https://cmake.org/cmake/help/latest/command/find_package.html I think setting this variable could help; https://cmake.org/cmake/help/latest/variable/CMAKE_FIND_ROOT_PATH.html#variable:CMAKE_FIND_ROOT_PATH

proppy:
> Reading thru: https://cmake.org/cmake/help/latest/command/find_package.html I think setting this variable could help; https://cmake.org/cmake/help/latest/variable/CMAKE_FIND_ROOT_PATH.html#variable:CMAKE_FIND_ROOT_PATH[](https://github.com/proppy)

That seems to work PTAL.

proppy:
Looks like the build completed successfully this time: https://github.com/hdl/conda-eda/runs/5270144212?check_suite_focus=true

proppy:
Curious if this should go in `pnr` (technically it's doing more than that), `misc` or a new category (`flow`)? @mithro @QuantamHD

ajelinski (changes requested):
Apart from replacing `boost` with `boost-cpp` (if possible), add `tk`, `zlib`, `libgomp`, `fmt` packages as `host` requirements.

After the package is built, its binaries' are always checked for dependencies and if all the libraries are provided by the `run` packages. Otherwise, such warnings are printed:
```
2022-02-21T09:05:38.4156357Z WARNING (openroad,bin/openroad): Needed DSO lib/libboost_serialization.so.1.76.0 found in ['boost-cpp']
2022-02-21T09:05:38.4157068Z WARNING (openroad,bin/openroad): .. but ['boost-cpp'] not in reqs/run, (i.e. it is overlinking) (likely) or a missing dependency (less likely)
```

Taking the `run-exports` mechanism into consideration, it most probably means some packages weren't used as `host` requirements even though their libraries are used during linking. Such packages could've been taken from the host OS or `build` reqs then. The latter is true for, e.g, the tcl library: https://github.com/hdl/conda-eda/runs/5270144212?check_suite_focus=true#step:3:1942

proppy:
> Otherwise, such warnings are printed:

Thanks for flagging, I somehow missed those. It would be neat to have a way to make `conda-build` treats `WARNING` as errors.

proppy:
We also get a bunch of:
```
WARNING :: Failed to get_static_lib_exports
```
but I'm not sure if they are fixable / should be ignored.

proppy:
PTAL, updated the host dependencies to fix the warnings.

pnr/openroad/meta.yaml
```diff
@@ -20,6 +20,10 @@ build:
   string: {{ environ.get('DATE_STR') }}
   script_env:
     - CI
+  ignore_run_exports_from:
+    # header-only libraries
+    - lemon
+    - fmt
```
ajelinski, line 26: Nice! :)

ajelinski (approved):
Conda-wise LGTM. I don't know about this `cudd` package but if you decide it's good to be merged please clean git history.

ajelinski:
> We also get a bunch of:
> 
> ```
> WARNING :: Failed to get_static_lib_exports
> ```
> 
> from https://github.com/conda/conda-build/blob/master/conda_build/os_utils/liefldd.py#L839 but I'm not sure if they are fixable / should be ignored.

They're common, I never considered them to be important.

proppy:
Rebased and squashed!

pnr/openroad/meta.yaml
proppy: It doesn't seems required per: https://github.com/The-OpenROAD-Project/OpenSTA/blob/2bc6e8f68ccff771559610413a8e2f57c40f9eff/README.md#build

> CUDD is a binary decision diageram (BDD) package that is used to improve conditional timing arc handling. OpenSTA does not require it to be installed.

PR #168, opened 2022-02-08
hls: add xls

### LINKED ISSUE #169: add xls package
upstream: https://github.com/google/xls
recipe: https://google.github.io/xls/tutorials/hello_xls/#1-installation-and-building
list of produced binaries: https://google.github.io/xls/tools/

proppy:
Fixes #169

- package the following binaries:
  - `interpreter_main`
  - `ir_converter_main`
  - `opt_main`
  - `codegen_main`
- drop python bazel deps in favor of conda

### DIFF
--- a/hls/xls/build.sh
+++ b/hls/xls/build.sh
@@ -1,0 +1,15 @@
+#!/bin/bash
+
+set -ex
+
+bazel build -c opt \
+  --extra_toolchains=@llvm_toolchain//:cc-toolchain-x86_64-linux \
+  //xls/dslx:interpreter_main \
+  //xls/dslx:ir_converter_main \
+  //xls/tools:opt_main \
+  //xls/tools:codegen_main
+
+cp bazel-bin/xls/dslx/interpreter_main \
+   bazel-bin/xls/dslx/ir_converter_main \
+   bazel-bin/xls/tools/opt_main \
+   bazel-bin/xls/tools/codegen_main $PREFIX/bin/
--- a/hls/xls/meta.yaml
+++ b/hls/xls/meta.yaml
@@ -1,0 +1,54 @@
+# Use `conda-build-prepare` before building for a better version string.
+{% set version = '%s_%04i_%s'| format(GIT_DESCRIBE_TAG | default('0.0.0') , GIT_DESCRIBE_NUMBER | default(0) | int, GIT_DESCRIBE_HASH | default('gUNKNOWN')) %}
+
+package:
+  name: xls
+  version: {{ version }}
+
+source:
+  - git_url: https://github.com/google/xls.git
+    git_rev: main
+    patches:
+      - skip-python-requirements.patch
+      - use-llvm-toolchain.patch      
+
+build:
+  # number: 202202082028
+  number: {{ environ.get('DATE_NUM') }}
+  # string: 20220208_2028
+  string: {{ environ.get('DATE_STR') }}
+  script_env:
+    - CI
+
+requirements:
+  build:
+    - bazel
+  host:
+    - python
+    - pip
+    - protobuf
+    - pycparser 2.19
+    - Flask 1.1.2
+    - Jinja2 2.11.3
+    - werkzeug 1.0.1
+    - itsdangerous 1.1.0
+    - MarkupSafe 2.0
+    - click 7.1.2
+    - numpy >=1.19.5,<=1.20.0
+    - scipy >=1.5.4,<=1.6.0
+    
+test:
+  commands:
+    - interpreter_main --version
+    - ir_converter_main --version
+    - opt_main --version
+    - codegen_main --version
+
+about:
+  home: https://google.github.io/xls/
+  license: Apache-2.0
+  summary: Accelerated HW Synthesis
+
+extra:
+  recipe-maintainers:
+    - proppy
--- a/.github/workflows/build-packages.yml
+++ b/.github/workflows/build-packages.yml
@@ -18,7 +18,7 @@ on:
 env:
   ANACONDA_TOKEN: ${{ secrets.ANACONDA_TOKEN }}
   ANACONDA_USER: ${{ secrets.ANACONDA_USER }}
-  NUM_OF_JOBS: 69
+  NUM_OF_JOBS: 70
 defaults:
   run:
     shell: bash
@@ -774,6 +774,18 @@ jobs:
     runs-on: "ubuntu-20.04"
     env:
       PACKAGE: "pnr/openroad"
+      OS_NAME: "linux"
+    steps:
+      - uses: actions/checkout@v2
+      - uses: hdl/conda-ci@master
+
+  #70
+  xls-linux:
+    # uses older ubuntu for better compat:
+    # https://github.com/hdl/conda-eda/issues/199
+    runs-on: "ubuntu-18.04"
+    env:
+      PACKAGE: "hls/xls"
       OS_NAME: "linux"
     steps:
       - uses: actions/checkout@v2
--- a/hls/xls/add-dist-rule.patch
+++ b/hls/xls/add-dist-rule.patch
@@ -1,0 +1,42 @@
+diff --git a/WORKSPACE b/WORKSPACE
+index 9bdffca5f..3466a8e36 100644
+--- a/WORKSPACE
++++ b/WORKSPACE
+@@ -50,3 +50,15 @@ initialize_external_repositories()
+ load("@com_github_grpc_grpc//bazel:grpc_extra_deps.bzl", "grpc_extra_deps")
+ 
+ grpc_extra_deps()
++
++http_archive(
++    name = "rules_pkg",
++    urls = [
++        "https://mirror.bazel.build/github.com/bazelbuild/rules_pkg/releases/download/0.5.1/rules_pkg-0.5.1.tar.gz",
++        "https://github.com/bazelbuild/rules_pkg/releases/download/0.5.1/rules_pkg-0.5.1.tar.gz",
++    ],
++    sha256 = "a89e203d3cf264e564fcb96b6e06dd70bc0557356eb48400ce4b5d97c2c3720d",
++)
++
++load("@rules_pkg//:deps.bzl", "rules_pkg_dependencies")
++rules_pkg_dependencies()
+diff --git a/xls/BUILD b/xls/BUILD
+index f97480caf..d6bd5292b 100644
+--- a/xls/BUILD
++++ b/xls/BUILD
+@@ -83,4 +83,17 @@ package(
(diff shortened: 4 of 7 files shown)


hls/xls/meta.yaml
```diff
@@ -0,0 +1,53 @@
...
+    - pip
+    - pycparser
+    - Flask
+    - Jinja2
+    - werkzeug
+    - itsdangerous
+    - click
+    - protobuf
+    - numpy
+    - scipy
```
ajelinski, line 37: Not sure why but this set of dependencies is a hell for Conda and rendering it for building takes unusually long time.
Shouldn't at least some of the versions match https://github.com/google/xls/blob/692e37534dfcd1af41b0b8b04ff2004d6a8193c5/dependency_support/pip_requirements.txt ?

I can see even setting a version for only `numpy` reduces the rendering time a lot so please set at least that one.

You can use `- numpy >=1.19.5,<=1.20` to have it the same as in the `requirements.txt` file or simply `- numpy 1.20` to have it use any `1.20.*` version.

hls/xls/meta.yaml
```diff
@@ -0,0 +1,53 @@
...
+  string: {{ environ.get('DATE_STR') }}
+  script_env:
+    - CI
+
+requirements:
+  build:
+    - bazel
+  host:
+    - python
+    - pip
```
ajelinski, line 28:
```suggestion
```
It isn't really necessary unless you want it to depend on a specific version.

hls/xls/meta.yaml
```diff
@@ -0,0 +1,53 @@
...
+    - pycparser
+    - Flask
+    - Jinja2
+    - werkzeug
+    - itsdangerous
+    - click
+    - protobuf
```
ajelinski, line 35: Is it correct? It's absent in the https://github.com/google/xls/blob/692e37534dfcd1af41b0b8b04ff2004d6a8193c5/dependency_support/pip_requirements.txt

ajelinski (changes requested):
There's also a problem with the recipe not using Conda compiler but a system one which links glibc, libstdc++ etc. that might be newer than the ones in the user's OS. That's why it should use Conda compiler package but I've tested that and the building fails.

They can be linked statically as a workaround with such a patch if that's an option?
```patch
diff --git a/xls/dslx/BUILD b/xls/dslx/BUILD
index 96fecc7b..3da3bdec 100644
--- a/xls/dslx/BUILD
+++ b/xls/dslx/BUILD
@@ -1285,6 +1285,7 @@ cc_binary(
     name = "ir_converter_main",
     srcs = ["ir_converter_main.cc"],
     visibility = ["//xls:xls_users"],
+    linkopts = ["-static"],
     deps = [
         ":command_line_utils",
         ":create_import_data",
@@ -1319,6 +1320,7 @@ cc_binary(
     name = "interpreter_main",
     srcs = ["interpreter_main.cc"],
     visibility = ["//xls:xls_users"],
+    linkopts = ["-static"],
     deps = [
         ":command_line_utils",
         ":error_printer",
diff --git a/xls/tools/BUILD b/xls/tools/BUILD
index aff5c380..2114a139 100644
--- a/xls/tools/BUILD
+++ b/xls/tools/BUILD
@@ -664,6 +664,7 @@ cc_binary(
     name = "opt_main",
     srcs = ["opt_main.cc"],
     visibility = ["//xls:xls_users"],
+    linkopts = ["-static"],
     deps = [
         ":opt",
         "@com_google_absl//absl/status",
@@ -826,6 +827,7 @@ cc_binary(
     name = "codegen_main",
     srcs = ["codegen_main.cc"],
     visibility = ["//xls:xls_users"],
+    linkopts = ["-static"],
     deps = [
         "@com_google_absl//absl/flags:flag",
         "@com_google_absl//absl/status",
```

Also, since in the package there are only executables, is Python required to use them? I'm wondering because with the current recipe it's going to be implicitly used as a `run` dependency (with the same major version) while the executables don't seem to need it, at least explicitly.

proppy:
>  There's a problem with the recipe not using Conda compiler but a system one which links glibc, libstdc++ etc.

Good point, I assumed that it would use the toolchain self-contained in the bazel `WORKSPACE` https://github.com/google/xls/blob/main/WORKSPACE#L24 but that doesn't seems to be the case (and I think we also wouldn't want that). I'll add a patch that also remove this out.

hls/xls/meta.yaml
```diff
@@ -0,0 +1,53 @@
...
+    - numpy
+    - scipy
+
+test:
+  commands:
+    interpreter_main --version
+    ir_converter_main --version
+    opt_main --version
+    codegen_main --version
```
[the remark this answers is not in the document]
proppy: good catch, I'm not sure why the test weren't failing with this (and what it was testing in the first place :)

hls/xls/build.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+set -ex
+
+bazel build -c opt //xls:dist
+
+tar -xvzf bazel-bin/xls/dist.tar.gz -C $PREFIX
```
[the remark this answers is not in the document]
proppy: oh you're right, I'm totally putting that in the wrong place ;)

Maybe we should use https://github.com/google/bazel_rules_install instead.

hls/xls/meta.yaml
proppy: I think that was needed because of https://github.com/google/xls/blob/629d348b6320d912206701afe064f28fa85ca948/xls/solvers/python/BUILD#L63

hls/xls/meta.yaml
proppy: you're right we should definitly have tighter version if it helps conda converge quicker, I'll see if we can get rid of the dependencies in the first place because we're not distributing the corresponding binary https://github.com/google/xls/blob/bca00e53bdbe0e4428f7d05009395602796402ff/xls/visualization/ir_viz/BUILD#L43

proppy:
>  That's why it should use Conda compiler package 

+1

> but I've tested that and the building fails

I also got a failure when it tried to compile llvm as a third_party dep with the conda compiler due to undefined `PRIXx32/64` macro, looks like we need a patch for LLVM to include the `<ctinttypes>` or patch the build system to link against conda-forge version? https://anaconda.org/conda-forge/llvm

ajelinski:
> >  That's why it should use Conda compiler package 
> 
> +1
> 
> > but I've tested that and the building fails
> 
> I also got a failure when it tried to compile llvm as a third_party dep with the conda compiler due to undefined `PRIXx32/64` macro, looks like we need a patch for LLVM to include the `<ctinttypes>` or patch the build system to link against conda-forge version? https://anaconda.org/conda-forge/llvm

Adding a patch for LLVM seems to be a good option.

QuantamHD:
@proppy To use the bazel self contained toolchain you need to add a flag to bazel build 
`--crosstool_top=@llvm_toolchain//:toolchain`

proppy:
@QuantamHD yes, but @ajelinski (and conda-build docs) suggest that the conda compilers should be used instead.

proppy:
rebased and squashed.

proppy:
actually went for @QuantamHD suggestion for now of using the embedded toolchain until we spend more time figure out how to build llvm using the conda gcc-based toolchain.

ajelinski:
I guess it's acceptable if there's no other option.
However, I've checked the binaries and they might not work on some older OSs (like Debian Buster or Ubuntu 18.04) because functions from a relatively recent `GLIBC_2.30` are required in `interpreter_main` and `ir_converter_main`.
Perhaps let's use `ubuntu-18.04` for this job in CI so the package is usable for more users.

Apart from this problem I have no further remarks except rebasing.

proppy:
> However, I've checked the binaries and they might not work on some older OSs (like Debian Buster or Ubuntu 18.04) because functions from a relatively recent GLIBC_2.30 are required in interpreter_main and ir_converter_main.

@ajelinski do you have an error at install time or at runtime?

ajelinski:
> > However, I've checked the binaries and they might not work on some older OSs (like Debian Buster or Ubuntu 18.04) because functions from a relatively recent GLIBC_2.30 are required in interpreter_main and ir_converter_main.
> 
> @ajelinski do you have an error at install time or at runtime?

There's no problem to install such a package but during runtime you will get an error.
For example a package built on Ubuntu 20.04 but installed and run on Ubuntu 18.04:
```bash
$ interpreter_main 
interpreter_main: /lib/x86_64-linux-gnu/libm.so.6: version `GLIBC_2.29' not found (required by interpreter_main)
interpreter_main: /lib/x86_64-linux-gnu/libpthread.so.0: version `GLIBC_2.30' not found (required by interpreter_main)
```

proppy:
> For example a package built on Ubuntu 20.04 but installed and run on Ubuntu 18.04:

Interesting I would have expected those to fail at install time when trying to resolve the `__glib` metapackage run_constraints, as documented here:
https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-virtual.html
https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#run-constrained

proppy:
rebased PTAL.

> let's use ubuntu-18.04 for this job in CI so the package is usable for more users

Isn't that something that would make sense to do for all packages though?

ajelinski:
> Interesting I would have expected those to fail at install time when trying to resolve the `__glib` metapackage run_constraints, as documented here: https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-virtual.html https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#run-constrained

I'm not really sure what these virtual packages do. I've only seen this @__glibc@ creating some unreasonable conflicts or at least Conda using it as an explanation when there was another underlying problem.

Typically when the Conda-packaged compiler is used it makes sure to link against a Conda-packaged glibc and adds a package with it as a dependency to have it installed with such a package. It's not the case here as a Conda-packaged compiler isn't used.

That's why typically we make sure it's used for package which is added.
Since it's hard to do here, we can handle the glibc versioning problem caring more about the OS version (hence glibc version) used to build the package.

BTW. I've never tried that but perhaps adding `__glibc 2.30` as a run requirement can trigger an error during the installation in such a case. It definitely isn't done automatically by `conda-build` though as after building the package doesn't depend on `__glibc` at all and I think we have better options here.

proppy:
@ajelinski PTAL, also update the workflow to use 18.04 and reference the issue you just filed ;)

ajelinski (approved):

ajelinski:
I can see a few Windows jobs have to be fixed so don't expect xls to be available after the CI finishes but I'm working on a fix.

PR #177, opened 2022-02-17
misc/open_pdks: add magic as optional deps

proppy:
This should ensure that if installed in the same envirnment, magic has
the same version as the one used to build open_pdks.

### DIFF
--- a/misc/open_pdks/meta.yaml
+++ b/misc/open_pdks/meta.yaml
@@ -38,6 +38,8 @@ requirements:
     - python
     - pip
     - magic
+  run_constrained:
+    - {{ pin_compatible('magic', min_pin='x.x.x', max_pin='x.x.x') }}
 
 test:
   commands:


misc/open_pdks/meta.yaml
```diff
@@ -38,6 +38,8 @@ requirements:
     - python
     - pip
     - magic
+  run_constrained:
+    - magic
```
ajelinski, line 42: You need to add some version specifier to make it constrained. The `pin_compatible` function comes handy here, you can change this line to:
```
    - {{ pin_compatible('magic', min_pin='x.x.x', max_pin='x.x.x') }}
```
and then the package will be constrained to run only with, e.g., `magic >=8.3.269_0_g7d60162,<8.3.270.0a0` if `magic 8.3.269_0_g7d60162` was used during building.

It's also possible to have the exact same package there (version + build string) with such a little hack I'm using for alias packages:
```
  run_constrained:
  {% for package in resolved_packages('host') %}
    {% if package.startswith( 'magic ' ) %}
    - {{ package }}
    {% endif %}
  {% endfor %}
```
and then it'll be only possible to install `magic 8.3.269_0_g7d60162 20220215_161633` but I think the `pin_compatible` option might be a better option here.

ajelinski (changes requested):

misc/open_pdks/meta.yaml
proppy: `pin_compatible` looks great! thanks for pointing it out! I was under the false impression that without any version it would just use the same one that was used for building (because that's what `conda render` was showing) but now I understand that you have to a be a little bit more explicit than that.

proppy: Done

proppy:
rebased, PTAL

proppy:
forgot to `git add` PTAL.

ajelinski (approved):

PR #185, opened 2022-02-24
misc/magic: add commit flag

proppy:
That'll help generating a valid SOURCES file for the open_pdks package in the format that OpenLane is expecting:
https://github.com/The-OpenROAD-Project/OpenLane/blob/master/docker/open_pdks/Dockerfile#L87

/cc @RTimothyEdwards let me know if you'd like this patch upstream.

### DIFF
--- a/misc/magic/add-commit-flag.patch
+++ b/misc/magic/add-commit-flag.patch
@@ -1,0 +1,51 @@
+diff --git a/scripts/defs.mak.in b/scripts/defs.mak.in
+index 77dfddc..1e74b17 100755
+--- a/scripts/defs.mak.in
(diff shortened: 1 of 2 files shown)


ajelinski (approved):

ajelinski:
I had no idea the changes have been upstreamed already but I've just removed the conflicting patch effectively leaving only the new test from this PR.

proppy:
Yep, it was upstreamed in https://github.com/RTimothyEdwards/magic/commit/bae5d9cd06c5999ad7952b38b0a4cbf676a9017b (@RTimothyEdwards happened to have independently worked on it!). Sorry for the lack of updates :)

PR #196, opened 2022-03-11
misc/netgen: remove conda-forge deps

proppy:
- remove conda-forge deps (tcsh)
- use cdt for x11 libs

Related to #193


ajelinski (approved):
Not sure if these `# [linux]` selectors are needed for CDT packages but keeping them shouldn't hurt.

PR #202, opened 2022-03-30
misc/magic: drop usptreamed patches

proppy:


ajelinski (approved):
`magic` is fixed and `open_pdks` can be successfully built with such a `magic` package so let's merge.

PR #216, opened 2022-05-27
ci: use a reusable workflow, instead of waiting and counting jobs 

umarcor:


ci/wait-for-statuses.py
```diff
@@ -49,13 +36,11 @@
...
-    subprocess.call(os.path.join(scripts_dir, "master-package.sh"))
-else:
+# Upload packages only when whole build succeeded and we are on master branch
+if not (branch == 'refs/heads/master'):
   print("Not on master branch, don't execute master-package.sh. Current branch: " + str(branch))
+elif(not jobFailure):
+  subprocess.call(os.path.join(scripts_dir, "master-package.sh"))
 
 # Always clean up
 subprocess.call(os.path.join(scripts_dir, "cleanup-anaconda.sh"))
```
ajelinski, line 46: This is a minor drawback of the change, the cleaning script will now only be run if all the packages were successfully built. Perhaps you could extract it to a separate job if it's not too much work otherwise just create an issue to remember about it and remove the comment above.

ajelinski (approved):
Generally LGTM, this approach is much better. It'd be great to have the cleanup script extracted to a separate job but it's not a blocker.

ci/wait-for-statuses.py
umarcor: I'm not sure I understand the proposed change. I did two modifications:

- Made the Upload job run unconditionally (`always()`).
- Cleaned the `wait-for-statuses` to execute `master-package.sh` regardless of job failures.

Let me know whether that's what you expected.

ajelinski:
Great! It's much cleaner this way.
Perhaps you could squash the commits before merging? Otherwise LGTM

PR #221, opened 2022-06-22
Fix icestorm and various yosys packages

PiotrZierhoffer:


PiotrZierhoffer:
Great work. Green again. There are some open questions here, but we'll address them one by one

PR #225, opened 2022-07-08
sim/verilator: linux: Require compiler package v8 or newer

ajelinski:
These changes should fix the Verilator's "Could not find compiler" issue: https://github.com/google/CFU-Playground/issues/568
For no good reason Conda installs a very old compiler package with Verilator even though there are no conflicts with the latest one: https://github.com/google/CFU-Playground/issues/568#issuecomment-1178890026

The old compiler packages (prior to v8) used a different prefix for the executables hence the problem.
Let's prevent Conda from installing Verilator with such compiler packages.

### DIFF
--- a/sim/verilator/meta.yaml
+++ b/sim/verilator/meta.yaml
@@ -20,10 +20,10 @@ build:
 
 requirements:
   build:
-    - {{ compiler('c') }}       [not osx]
-    - {{ compiler('cxx') }}     [not osx]
-    - {{ compiler('c') }} 4.0   [osx]
-    - {{ compiler('cxx') }} 4.0 [osx]
+    # Older Linux compiler packages doesn't contain the '...-conda-...'-prefixed
+    #   executables which are common since v8. They're required for verilation.
+    - {{ compiler('cxx') }}
+    - gxx_linux-64 >=8      [linux]
     - autoconf
     - make
   host:
@@ -32,6 +32,7 @@ requirements:
     - zlib
   run:
     - {{ compiler('cxx') }}
+    - gxx_linux-64 >=8      [linux]
     - bison
     - flex
     - ncurses
--- a/sim/verilator/meta.yaml
+++ b/sim/verilator/meta.yaml
@@ -38,8 +38,6 @@ requirements:
 
 test:
   requires:
-    - gxx_linux-64   [linux]
-    - clangxx_osx-64 [osx]
     - make
   files:
     - test/counter.v


PiotrZierhoffer (approved):
Lgtm if green!

ajelinski:
Verilator green so I'm merging.

tcal-x:
Hi @ajelinski ,
I tried using the updated package in CFU Playground, and encountered some other issues when it came to actually linking the simulation executable.   I added a link test to the Verilator testcase repo to show the behavior, and added comments to https://github.com/tcal-x/verilator-make-env-test/issues/1.

ajelinski:
Hi @tcal-x,
I'll answer here in case anyone else has the same problem in the future as I'm not sure if you're going to keep the `-test` repository around.

> After the recent update by Adam, I no longer see the issue with the compiler not found.
> 
> However, there are issues when it comes to linking the Verilated simulation with testbench. To illustrate, I have added a `make link` target in the `test/` directory Makefile.
> 
> When I run this, I get:
> 
> ```
> ...
> echo "" > Vadd__ALL.verilator_deplist.tmp
> Archive x86_64-conda-linux-gnu-ar -rcs Vadd__ALL.a Vadd__ALL.o
> x86_64-conda-linux-gnu-c++ -Wl,-O2 -Wl,--sort-common -Wl,--as-needed -Wl,-z,relro -Wl,-z,now -Wl,--disable-new-dtags -Wl,--gc-sections -Wl,-rpath,/home/tim/tcal-x/verilator-make-env-test/env/conda/envs/verilator-make-env/lib -Wl,-rpath-link,/home/tim/tcal-x/verilator-make-env-test/env/conda/envs/verilator-make-env/lib -L/home/tim/tcal-x/verilator-make-env-test/env/conda/envs/verilator-make-env/lib    tb.o verilated.o Vadd__ALL.a      -o Vadd
> /home/tim/tcal-x/verilator-make-env-test/env/conda/envs/verilator-make-env/bin/../lib/gcc/x86_64-conda-linux-gnu/11.2.0/../../../../x86_64-conda-linux-gnu/bin/ld: /home/tim/tcal-x/verilator-make-env-test/env/conda/envs/verilator-make-env/lib/libstdc++.so: undefined reference to `memcpy@GLIBC_2.14'
> /home/tim/tcal-x/verilator-make-env-test/env/conda/envs/verilator-make-env/bin/../lib/gcc/x86_64-conda-linux-gnu/11.2.0/../../../../x86_64-conda-linux-gnu/bin/ld: /home/tim/tcal-x/verilator-make-env-test/env/conda/envs/verilator-make-env/lib/libstdc++.so: undefined reference to `aligned_alloc@GLIBC_2.16'
> /home/tim/tcal-x/verilator-make-env-test/env/conda/envs/verilator-make-env/bin/../lib/gcc/x86_64-conda-linux-gnu/11.2.0/../../../../x86_64-conda-linux-gnu/bin/ld: /home/tim/tcal-x/verilator-make-env-test/env/conda/envs/verilator-make-env/lib/libstdc++.so: undefined reference to `clock_gettime@GLIBC_2.17'
> collect2: error: ld returned 1 exit status
> make[3]: *** [Vadd.mk:65: Vadd] Error 1
> rm Vadd__ALL.verilator_deplist.tmp
> ```

It turns out Conda fails to use `libc-2.17.so` provided for each environment through the `sysroot_linux-64-2.17` package and uses glibc from the host OS for running (it seems to work properly for package building though)...

I've never noticed it TBH, using Debian I'd need to use Debian 7 (jessie) to have `libc` older than that. That's why the `make link` works for me instantly in a fresh `debian:bullseye` container.

I've checked the `libstdcxx-ng` packages which provide the `libstdc++` library in Conda (https://anaconda.org/main/libstdcxx-ng/files) and found out all the packages built in July 2021 and newer have been linked against the `libc>2.17`. There are, however, v9 and v8 packages built late 2019.

Therefore, as a workaround to solve the problem on OSs with `libc` older than v2.17, you could use `libstdcxx-ng=9.1` or `libstdcxx-ng=8.2` package. v7.3 is to be avoided because then C++ compiler <v8 will be installed and the original problem will occur again.

tcal-x:
@ajelinski , to be clear, you're suggesting that I add a line
```
  - libstdcxx-ng=9.1
```
to *my* `environment.yml` files(s), either in the test repo, or in CFU Playground?

When I do that, I still have one complaint from the linker (1 out of the original 3):
```
/home/tim/tcal-x/verilator-make-env-test/env/conda/envs/verilator-make-env/bin/../lib/gcc/x86_64-conda_cos6-linux-gnu/7.3.0/../../../../x86_64-conda_cos6-linux-gnu/bin/ld: /home/tim/tcal-x/verilator-make-env-test/env/conda/envs/verilator-make-env/lib/libgcc_s.so.1: undefined reference to `memcpy@GLIBC_2.14'
```

ajelinski:
Yes, I suggested adding such a line to the `environment.yml` wherever this problem occurs.

I was mistaken Conda fails to use the provided `libc-2.17.so` because it seems it's only meant to be used for the package building. Since it's used during building, the host OS's `libc` should also be at least v2.17 for the Conda packages to work correctly. This version seems to have been used for the package building for the last 17 months: https://anaconda.org/main/sysroot_linux-64/files

This time the problem is about `libgcc_s.so.1`. Using a `libgcc-ng=9.3` (it might be 9.1 as well) restriction should help for this particular error but it might just be the tip of the iceberg.

I've also noticed the `libstdcxx-ng=9.1` restriction won't allow any Verilator newer than v4.202 to be installed because since v4.204 the Verilator packages were linked against `libstdc++` v9.3...

So, TBH, using the host OS with `libc` >=2.17 is the real solution to the linking problem.
Please let me know if that's problematic with the colabs.

mithro:
The conda packages should not be using anything from the host.

tcal-x:
@ajelinski adding `libgcc-ng=9.1` (and removing `libstdcxx-ng=9.1` -- not needed, but doesn't hurt either) allows me to link successfully in the test repo.   I'll try now in CFU Playground.

tcal-x:
No success...with `libgcc-ng=9.1` in CFU Playground, there are unresolved conflicts with other packages.

Some of the stuff printed out regarding the conflicts seems suspicious, but I'm not an expert, so maybe it's normal.    Verilator required one specific version of `flex` for example.

ajelinski:
> The conda packages should not be using anything from the host.

It turns out `libc` is always used from the host OS. I didn't know that either. Even Python is linked against the glibc libraries but there are no `libc`/`libm` etc. within the Conda tree, e.g., after using the Miniconda installer.

I'm kinda surprised you had no problems earlier @tcal-x. I've just checked and, e.g., an [11-month-old Python 3.7.11](https://anaconda.org/main/python/3.7.11/download/linux-64/python-3.7.11-h12debd9_0.tar.bz2) already requires `libc` to be at least v2.17 (based on the `readelf -V` output).

The `libc-2.17.so` library is only provided with the compiler packages to make sure no Conda package requires any newer `libc` than that but that's it. I guess Conda guys decided it shouldn't be problematic since it was released almost 10 years ago. For example, I see VMs for Colab offer much newer `libc-2.27.so` even though the OS used isn't any of the latest ones (Ubuntu 18.04).

tcal-x:
FWIW, when I add these lines (which make things work in the test repo):
```
 - libgcc-ng=9.1
```
...I get this error from the Conda install:
```
The following specifications were found to be incompatible with your system:

  - feature:/linux-64::__glibc==2.31=0
  - feature:|@/linux-64::__glibc==2.31=0
  - litex-hub::dfu-util -> libusb -> __glibc[version='>=2.17|>=2.17,<3.0.a0']
  - litex-hub::gcc-riscv32-elf-newlib -> libgcc-ng[version='>=9.3.0'] -> __glibc[version='>=2.17']
  - litex-hub::iceprog -> libgcc-ng[version='>=9.3.0'] -> __glibc[version='>=2.17']
  - litex-hub::nextpnr-ecp5 -> libgcc-ng[version='>=9.3.0'] -> __glibc[version='>=2.17']
  - litex-hub::nextpnr-ice40 -> libgcc-ng[version='>=9.3.0'] -> __glibc[version='>=2.17']
  - litex-hub::nextpnr-nexus -> libgcc-ng[version='>=9.3.0'] -> __glibc[version='>=2.17']
  - litex-hub::openocd -> libusb -> __glibc[version='>=2.17|>=2.17,<3.0.a0']
  - litex-hub::symbiflow-yosys-plugins -> libgcc-ng[version='>=9.3.0'] -> __glibc[version='>=2.17']
  - litex-hub::yosys -> libgcc-ng[version='>=9.3.0'] -> __glibc[version='>=2.17']
  - python=3.7 -> libgcc-ng[version='>=7.5.0'] -> __glibc[version='>=2.17']

Your installed version is: 2.31
```
I'm not sure what the conflict is, since all of the constraints seem to be `>=2.17`.

PR #226, opened 2022-07-20
Fix compiler for yosys and plugins builds

ajelinski:
Fixes for two problems caused by my previous changes (https://github.com/hdl/conda-eda/commit/2f65e11ca106a90ce2a664186a2019dea020c595 and https://github.com/hdl/conda-eda/commit/14bbf08faf8a67406fa9e1e46eab73a5e162ea81 respectively):
1. `yosys-config --cxx` returning the compiler's full path instead of just a name,
2. plugins built with the host's `g++`, therefore linked against a non-Conda `glibc`; it made running impossible on older OSs: 
    https://github.com/chipsalliance/f4pga-examples/runs/7392390070?check_suite_focus=true#step:7:37
    ```
    ERROR: Can't load module `./systemverilog': /lib/x86_64-linux-gnu/libm.so.6: version `GLIBC_2.29' not found (required by /root/opt/f4pga/xc7/conda/envs/xc7/bin/../share/yosys/plugins/systemverilog.so)
    ```


PiotrZierhoffer:
I'm generally ok with these. It's difficult to analyze just from the text, but if it solves the issue and gives us green ci - lgtm

ajelinski:
A proper compiler is now used to build plugins: https://github.com/hdl/conda-eda/runs/7427145728?check_suite_focus=true#step:3:1549
The result of `yosys-config --cxx` is also correct now.

The `tree-sitter-verilog` failure isn't caused by the PR changes so I'm merging.

PR #229, opened 2022-08-03
misc/tree-sitter-verilog: Use `py-tree-sitter` directly

ajelinski:
For as long as I can check in the CI logs there have been troubles with
installing the latest `tree_sitter` package from PyPI (v0.20 currently):
```
No matching distribution found for setuptools>=43.0.0
```

The job still worked because Pip was then checking whether older
packages work and installed the old `tree_sitter` v0.2.1 package.

The latest Pip v22.1.2 exits with failure instead which made the job
fail for a week or so.

By using `py-tree-sitter` directly always the latest version will be
used and the PyPI dependency is removed completely.

### DIFF
--- a/misc/tree-sitter-verilog/build.sh
+++ b/misc/tree-sitter-verilog/build.sh
@@ -7,6 +7,9 @@ npm install tree-sitter-cli
 
 ./node_modules/tree-sitter-cli/tree-sitter generate
 
-$PYTHON -m pip install --isolated tree_sitter
-$PYTHON -c "from tree_sitter import Language; Language.build_library(\"$PREFIX/lib/tree-sitter-verilog.so\", [\".\"])"
-$PYTHON -m pip uninstall -y tree_sitter
+# Build the 'tree_sitter/binding' extension.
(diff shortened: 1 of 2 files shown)


ajelinski:
Example for how pip previously worked can be found in the last successful job which used the old pip package, i.e., https://github.com/hdl/conda-eda/runs/7273911376
```
  Collecting tree_sitter
    Downloading tree_sitter-0.20.0.tar.gz (116 kB)
    Installing build dependencies: started
    Installing build dependencies: finished with status 'error'
    Downloading tree_sitter-0.19.0.tar.gz (112 kB)
    Installing build dependencies: started
    Installing build dependencies: finished with status 'error'
    Downloading tree_sitter-0.2.2.tar.gz (110 kB)
    Installing build dependencies: started
    Installing build dependencies: finished with status 'error'
    Downloading tree_sitter-0.2.1.tar.gz (110 kB)
(...)
  Successfully installed tree-sitter-0.2.1
```

PiotrZierhoffer (approved):
lgtm

PR #236, opened 2022-09-01
update conda before installing packages

apokusinski:
Installing python without updating conda beforehand caused an error `OSError: Could not find a suitable TLS CA certificate bundle, invalid path: /tmp/really-long-path/conda/lib/python3.9/site-packages/certifi/cacert.pem` in github actions. Updating conda before the installation fixed this issue.


PiotrZierhoffer (approved):
If we're green, then lgtm

PR #237, opened 2022-09-06
Try to explicitly force Conda includes

PiotrZierhoffer:


PiotrZierhoffer:
This approach didn't really help.

Meanwhile I removed stale quicklogic-* packages (upstream is used now) and fixed prjxray-tools and verible builds.

Windows seems to fail on git clone - might be temporary.

Nextpnr failes on using the host python instead of the one from conda.

PR #241, opened 2022-09-20
nextpnr/prjtrellis: Fix CMake find strategy for Python

ajelinski:
By default, CMake looks for the latest possible version of Python in the projects where CMake minimum version is less than 3.15.
Since 3.15 there's a new policy making it use the first Python found within the given version constraints.

The `nextpnr` and `prjtrellis` CMakeLists.txt allow CMake to be older than 3.15 which is why the old policy is used.
Forcing CMake to use a new policy fixes the nextpnr and prjtrellis problems with finding Python development headers on Linux.
After such a change CMake uses Conda Python which provides the headers instead of the host Python for which the headers aren't available.

There are some additional minor changes which are described in the commit messages.

### DIFF
--- a/bit/prjtrellis/build.sh
+++ b/bit/prjtrellis/build.sh
@@ -3,8 +3,18 @@
 set -e
 set -x
 
-cd libtrellis
+# Use 'Python3_FIND_STRATEGY=LOCATION' in projects with 'cmake_minimum_required'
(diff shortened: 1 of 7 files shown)


PiotrZierhoffer:
This fixes packages in question, merging

PR #243, opened 2022-09-21
nextpnr: Fix building on macOS

PiotrZierhoffer:
The current code uses features unavailable in older macOS.

If we're about to use this code, we need to enable them anyway. Conda has a define controlling it.


PiotrZierhoffer:
This was actually solved in #242, closing

PR #246, opened 2022-09-27
misc: add openlane

### LINKED ISSUE #244: add OpenLane package
# Upstream
https://github.com/The-OpenROAD-Project/OpenLane/

# License
https://github.com/The-OpenROAD-Project/OpenLane/blob/master/LICENSE

# Dependencies
- OpenROAD
- netgen
- open_pdks (optional?)
- magic
- pyyaml
- click
- tcllib (system?)
- yosys

# Test
```
flow.tcl -design inverter
```

proppy:
Fixes #244

/cc @QuantamHD @mithro

### DIFF
--- a/.github/workflows/Build.yml
+++ b/.github/workflows/Build.yml
@@ -719,3 +719,14 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - uses: ./ci
+
+  #64
+  openlane-linux:
+    needs: ["magic-linux", "netgen-linux", "openroad-linux", "yosys-linux-py37", "yosys-linux-py38"]
+    runs-on: "ubuntu-20.04"
+    env:
+      PACKAGE: "misc/openlane"
+      OS_NAME: "linux"
+    steps:
+      - uses: actions/checkout@v3
+      - uses: ./ci
--- a/misc/openlane/1439.patch
+++ b/misc/openlane/1439.patch
@@ -1,4 +1,4 @@
-From e57dc500ef22047e2d57c1c4bcae83260f6d3672 Mon Sep 17 00:00:00 2001
+# From e57dc500ef22047e2d57c1c4bcae83260f6d3672 Mon Sep 17 00:00:00 2001
 From: Johan Euphrosine <proppy@google.com>
 Date: Tue, 18 Oct 2022 22:29:37 +0900
 Subject: [PATCH] flow: skip mounted version for local install
--- a/misc/openlane/1439.patch
+++ b/misc/openlane/1439.patch
@@ -1,0 +1,29 @@
+From e57dc500ef22047e2d57c1c4bcae83260f6d3672 Mon Sep 17 00:00:00 2001
+From: Johan Euphrosine <proppy@google.com>
+Date: Tue, 18 Oct 2022 22:29:37 +0900
+Subject: [PATCH] flow: skip mounted version for local install
+
+Signed-off-by: Johan Euphrosine <proppy@google.com>
+---
+ flow.tcl | 8 +++++---
+ 1 file changed, 5 insertions(+), 3 deletions(-)
+
+diff --git a/flow.tcl b/flow.tcl
+index 2ec16863a..6a9eec912 100755
+--- a/flow.tcl
++++ b/flow.tcl
+@@ -386,9 +386,11 @@ if {[catch {exec cat $::env(OPENLANE_ROOT)/install/installed_version} ::env(OPEN
+     }
+ }
+ 
+-if {![catch {exec git --git-dir $::env(OPENLANE_ROOT)/.git rev-parse HEAD} ::env(OPENLANE_MOUNTED_SCRIPTS_VERSION)]} {
+-    if { $::env(OPENLANE_VERSION) == $::env(OPENLANE_MOUNTED_SCRIPTS_VERSION)} {
+-        unset ::env(OPENLANE_MOUNTED_SCRIPTS_VERSION)
++if { ! [info exists ::env(OPENLANE_LOCAL_INSTALL)] || ! $::env(OPENLANE_LOCAL_INSTALL)} {
++    if {![catch {exec git --git-dir $::env(OPENLANE_ROOT)/.git rev-parse HEAD} ::env(OPENLANE_MOUNTED_SCRIPTS_VERSION)]} {
++        if { $::env(OPENLANE_VERSION) == $::env(OPENLANE_MOUNTED_SCRIPTS_VERSION)} {
++            unset ::env(OPENLANE_MOUNTED_SCRIPTS_VERSION)
++        }
+     }
+ }
+ 
--- a/misc/openlane/build.sh
+++ b/misc/openlane/build.sh
@@ -1,0 +1,46 @@
+#!/bin/bash
+
+# Copyright 2022 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -ex
+
+# Identify OS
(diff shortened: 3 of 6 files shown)


.github/workflows/Build.yml
```diff
@@ -717,3 +717,17 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - uses: ./ci
+
+  #64
+  openlane-linux:
+    runs-on: "ubuntu-20.04"
```
ajelinski, line 723:
```suggestion
    needs: ["magic-linux", "netgen-linux", "openroad-linux", "yosys-linux-py37", "yosys-linux-py38"]
    runs-on: "ubuntu-20.04"
```

ajelinski: Perhaps let's wait for these jobs to finish so that the packages built in the same CI can be used during testing.

ajelinski (approved):
Apart from `needs` I see such logs:
* https://github.com/hdl/conda-eda/actions/runs/3136689055/jobs/5093922644#step:4:1726
  ```
    OpenLane UNKNOWN
    (with mounted scripts from fatal: not a git repository: '$PREFIX/share/openlane/.git')
    All rights reserved. (c) 2020-2022 Efabless Corporation and contributors.
    Available under the Apache License, version 2.0. See the LICENSE file for more details.
    
    Warning: : OpenLane may not function properly: not enough values to unpack (expected 3, got 1)
  ```
* https://github.com/hdl/conda-eda/actions/runs/3136689055/jobs/5093922644#step:4:1809
  ```
  .../openlane/designs/inverter/runs/RUN_2022.09.27_15.06.12/results/signoff/inverter.klayout.gds' wasn't found. Skipping GDS XOR.
  ```

Did you see them, are those acceptable? Otherwise LGTM.

ajelinski:
> Maybe we could patch https://github.com/The-OpenROAD-Project/OpenLane/blob/32da932761213af689f10088d671e1e3dc38f273/flow.tcl#L395-L398 so that it include the conda version?

Perhaps adding `echo $PKG_VERSION >$PREFIX/share/openlane/install/installed_version` to `build.sh` could be enough thanks to: https://github.com/The-OpenROAD-Project/OpenLane/blob/32da932761213af689f10088d671e1e3dc38f273/flow.tcl#L381

proppy:
> Perhaps adding echo $PKG_VERSION >$PREFIX/share/openlane/install/installed_version to build.sh could be

Yes figured this out too with https://github.com/hdl/conda-eda/pull/246/commits/ac1610246abca451b540dc649fa9186876c6a266 but I think we also need https://github.com/The-OpenROAD-Project/OpenLane/pull/1439.

proppy:
@ajelinski what do you think of bundling tcllib like I'm doing in https://github.com/proppy/conda-eda/tree/master/misc/tcllib ? would you prefer to have this as a separate PR?

ajelinski:
> @ajelinski what do you think of bundling tcllib like I'm doing in https://github.com/proppy/conda-eda/tree/master/misc/tcllib ? would you prefer to have this as a separate PR?

I have no problem with including it in this PR as such.

I don't know tcl very well, is tcllib OS-independent? Shouldn't it be actually built with `conda-build`?

proppy:
> I don't know tcl very well, is tcllib OS-independent? Shouldn't it be actually built with conda-build?

There is actually a conda-forge package we could backport: https://github.com/conda-forge/tcllib-feedstock

Looking at the underlying installer it seems that it's mostly copying files around: https://core.tcl-lang.org/tcllib/file?name=installer.tcl&ci=tip so I'm not sure why it's not noarch https://anaconda.org/conda-forge/tcllib/files

ajelinski:
> There is actually a conda-forge package we could backport: https://github.com/conda-forge/tcllib-feedstock
> 
> Looking at the underlying installer it seems that it's mostly copying files around: https://core.tcl-lang.org/tcllib/file?name=installer.tcl&ci=tip so I'm not sure why it's not noarch https://anaconda.org/conda-forge/tcllib/files

I'll "build" and upload the `tcllib` package from a branch then and move the package on the `litex-hub` channel to `main`. IMO adding it to the repository to have it built nightly doesn't make sense in such a case especially that the source doesn't change at all.

Perhaps it isn't `noarch` because it has the `tk` package in `run` requirements? Not sure but maybe let's keep it as it is especially that we only need the package for Linux, WDYT?

proppy:
> I'll "build" and upload the tcllib package from a branch then and move the package on the litex-hub channel to main. 

Oh I didn't realize you had upload write to anaconda main! I'll keep this in mind for other commonly used package that are only in `conda-forge`.

> Perhaps it isn't noarch because it has the tk package in run requirements? 

Ah that would make sense, looking at the various module documentation it seems that only only few package depends on Tk:
https://github.com/tcltk/tcllib/search?q=%22%5Brequire+Tk%22+titledesc:
- map
- comm
- tepam
- math

Looking at the installer it seems that there is a `+excluded` flag https://github.com/tcltk/tcllib/blob/efe4404da89b4bd7adafeadd4106a1a4592492ab/installer.tcl#L474 we could leverage to remove them (or we could patch out the part that depends on tk). I'm not familiar with the `main` requirement and that might be moot since `tk` looks already well packaged: https://anaconda.org/main/tk

ajelinski:
> Oh I didn't realize you had upload write to anaconda main! I'll keep this in mind for other commonly used package that are only in `conda-forge`.

Oh no, I'm not that powerful. :-) I only meant moving the package to the `main` Conda label. It only makes the package built by CI from a `conda-eda` branch available within the LiteX-Hub channel without providing the `ci-<branch>-<run_id>` label (because `main` is the default label).

> Ah that would make sense, looking at the various module documentation it seems that only only few package depends on Tk: https://github.com/tcltk/tcllib/search?q=%22%5Brequire+Tk%22+titledesc:
> 
>     * map
> 
>     * comm
> 
>     * tepam
> 
>     * math
> 
> 
> Looking at the installer it seems that there is a `+excluded` flag https://github.com/tcltk/tcllib/blob/efe4404da89b4bd7adafeadd4106a1a4592492ab/installer.tcl#L474 we could leverage to remove them (or we could patch out the part that depends on tk). I'm not familiar with the `main` requirement and that might be moot since `tk` looks already well packaged: https://anaconda.org/main/tk

Is having a `noarch` package worth such a struggle?

I've just made the `tcllib` package for Linux available on the LiteX-Hub channel so please test it and modify this package to use it.

proppy:
> Is having a noarch package worth such a struggle?

Not really, but it also allow us to remove the transitive dependencies on `tk` for openlane :)

proppy:
> I've just made the tcllib package for Linux available on the LiteX-Hub channel so please test it and modify this package to use it.

Done.

proppy:
> Not really, but it also allow us to remove the transitive dependencies on tk for openlane :)

Filed https://github.com/hdl/conda-eda/issues/250#issuecomment-1284136411 so that we can explore this separatly.

.github/workflows/Build.yml
```diff
@@ -717,3 +717,20 @@ jobs:
...
+      PACKAGE: "misc/openlane"
+      OS_NAME: "linux"
+    steps:
+      - uses: actions/checkout@v3
+      - name: 'Install tcllib'
+        run: |
+          sudo apt update
+          sudo apt install -y tcllib
+          TCLLIBPATH=/usr/share/tcltk/tcllib*
+          echo "TCLLIBPATH=$TCLLIBPATH" >> $GITHUB_ENV
```
ajelinski, line 735:
```suggestion
```

ajelinski: Please fix this and squash the commits.

ajelinski:
Unfortunately the build failed -- there's something wrong with the patch @proppy

proppy:
>  Unfortunately the build failed -- there's something wrong with the patch @proppy

refreshed the patch

misc/openlane/1439.patch
```diff
@@ -0,0 +1,29 @@
+From e57dc500ef22047e2d57c1c4bcae83260f6d3672 Mon Sep 17 00:00:00 2001
```
ajelinski, line 1:
```suggestion
# From e57dc500ef22047e2d57c1c4bcae83260f6d3672 Mon Sep 17 00:00:00 2001
```
Currently Conda-build fails to apply the patch because of the header. Please add a hash or even just a space at the beginning of this line to solve the problem.

proppy:
@ajelinski curious why the CI still fails?

ajelinski:
The patch was merged upstream yesterday. I've just merged and removed the patch. :)

Issue #204, opened 2022-04-18
mirror packages on google cloud storage


proppy:
Since conda has the capability of installing package from arbitrary static http host (as long as it can fetch the index), we could consider mirroring the packages under a google cloud storage bucket.

That would have the following advantages:
- no more arbitrary quota limitation
- faster fetch from GCP infra (incl dedicated runners #184)
- more relevant package name (we could use the `https://conda.hdl.tools` as the canonical channel name

proppy:
depending if we want to land this before or after #184, we could use either https://github.com/google-github-actions/upload-cloud-storage#authenticating-via-workload-identity-federation or https://github.com/google-github-actions/upload-cloud-storage#via-application-default-credentials for authorizing the action to access the bucket.

proppy:
@umarcor as discussed with @mithro on https://github.com/hdl/conda-eda/issues/184#issuecomment-1052923605, do you think we can get a bucket set-up on `github-hdl` with either https://cloud.google.com/iam/docs/workload-identity-federation or https://cloud.google.com/iam/docs/creating-managing-service-accounts authorization scheme set so that we can upload artefact from the `Upload packages` workflow?

proppy:
@PiotrZierhoffer @ajelinski @AdamOlech can we get a bucket created in the same project that's currently running the runners?

PR #247, opened 2022-09-27
add constructor based installer

proppy:
Depends on #246 

/cc @QuantamHD @umarcor @mithro

### DIFF
--- a/.github/workflows/Construct.yml
+++ b/.github/workflows/Construct.yml
@@ -1,0 +1,49 @@
+name: Build installers
+
+on:
+  workflow_call:
+
+permissions:
+  contents: write    
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  openlane-sky130a-installer:
+    runs-on: "ubuntu-20.04"
+    steps:
+      - name: checkout conda-eda
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0 # fetch all tags
+
+      - name: construct installer
+        run: |
+          export CI_SCRIPTS_PATH="$(pwd)/ci"
+          ./ci/install.sh
+          conda install constructor
+          constructor installer/openlane.sky130a/
+
+      - name: archive openlane.sky130a installer
+        uses: actions/upload-artifact@v3
+        with:
+          name: "openlane-sky130a-installer"
+          path: |
+            /home/runner/work/conda-eda/conda-eda/openlane-sky130a-0-Linux-x86_64.sh
+
+      - run: |
+          git config --local user.name conda-eda@bot
+          git config --local user.email conda-eda@bot
+          TAG=$(git describe --tags)
+          git tag $TAG
+          git push origin "HEAD:refs/tags/$TAG"
+          echo "TAG=$TAG" >> $GITHUB_ENV
+
+      - uses: softprops/action-gh-release@v1
+        with:
+          name: ${{ env.TAG }}
+          tag_name: ${{ env.TAG }}
+          files: |
+            openlane-sky130a-installer/*.sh
--- a/installer/openlane.sky130a/construct.yaml
+++ b/installer/openlane.sky130a/construct.yaml
@@ -1,0 +1,15 @@
+name: openlane-sky130a
+version: 0
+
+channels:
+  - https://conda.anaconda.org/litex-hub
+  - http://repo.anaconda.com/pkgs/main/
+
+specs:
+  - python 3.7*
+  - conda
+  - openlane
+  - open_pdks.sky130a
+  - xls
+
+post_install: post_install.sh
--- a/.github/workflows/Build.yml
+++ b/.github/workflows/Build.yml
@@ -717,3 +717,17 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - uses: ./ci
+
+  #64
+  openlane-linux:
+    runs-on: "ubuntu-20.04"
+    env:
+      PACKAGE: "misc/openlane"
+      OS_NAME: "linux"
+    steps:
+      - uses: actions/checkout@v3
+      - name: 'Install tcllib'
+        run: |
+          sudo apt update
+          sudo apt install -y tcllib
+      - uses: ./ci
--- a/.github/workflows/Upload.yml
+++ b/.github/workflows/Upload.yml
@@ -47,6 +47,12 @@ jobs:
           ANACONDA_USER: ${{ secrets.ANACONDA_USER }}
 
 
+  Construct:
+    needs: Upload
+    if: always() && (github.event_name != 'pull_request') && (github.ref == 'refs/heads/master')
+    uses: ./.github/workflows/Construct.yml
+
+
   Cleanup:
     runs-on: ubuntu-20.04
     steps:
--- a/installer/openlane.sky130a/post_install.sh
+++ b/installer/openlane.sky130a/post_install.sh
@@ -1,0 +1,9 @@
+#!/bin/bash
+
+set -ex
+
+cat >> $PREFIX/share/openlane/install/env.tcl <<EOF
+set ::env(PDK) "sky130A"
+set ::env(STD_CELL_LIBRARY) "sky130_fd_sc_hd"
+set ::env(STD_CELL_LIBRARY_OPT) "sky130_fd_sc_hd"
+EOF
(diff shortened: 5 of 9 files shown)


mithro:
@PiotrZierhoffer  / @umarcor - Could you please review? I think it looks fine and thus we should merge but are unsure about the failures?

mithro:
@proppy - Any idea why things are failing in the CI?

proppy:
seems unrelated, nightly CI has similar failure: https://github.com/hdl/conda-eda/actions/runs/3253556045

proppy:
but we can get https://github.com/hdl/conda-eda/pull/246 reviewed indepently if it eases the process.

.github/workflows/Construct.yml
```diff
@@ -0,0 +1,49 @@
...
+          ./ci/install.sh
+          conda install constructor
+          constructor installer/openlane.sky130a/
+
+      - name: archive openlane.sky130a installer
+        uses: actions/upload-artifact@v3
+        with:
+          name: "openlane-sky130a-installer"
+          path: |
+            /home/runner/work/conda-eda/conda-eda/openlane-sky130a-0-Linux-x86_64.sh
```
ajelinski, line 34: Isn't there any variable available with this path?

ajelinski (commented):
I'm genuinely impressed by the whole idea and Conda construct itself as I didn't know the tool.

I tested using the installer and unfortunately stumbled upon such an error:
```
error copying "/root/silicon-env/share/openlane/designs/inverter/runs/RUN_2022.10.18_13.16.56/results/signoff/inverter.magic.gds": no such file or directory
    while executing
"file copy -force $::env(MAGIC_GDS) $::env(CURRENT_GDS)"
    (procedure "run_magic" line 21)
    invoked from within
"run_magic"
    (procedure "run_magic_step" line 3)
    invoked from within
"[lindex $step_exe 0] [lindex $step_exe 1] "
    (procedure "run_non_interactive_mode" line 52)
    invoked from within
"run_non_interactive_mode {*}$argv"
    (file "/root/silicon-env/share/openlane/flow.tcl" line 415)
    invoked from within
"source "$::env(CONDA_PREFIX)/share/openlane/flow.tcl""
    (file "/root/silicon-env/bin/flow.tcl" line 3)
```
but perhaps it'll get fixed by disabling this klayout xor?

What bothers me a little is that we'll have a tag added for each run of the `Construct` workflow, i.e., with each `Upload` workflow run from master. Do I get it right? Isn't there any other option to publish the installer?

proppy:
> I tested using the installer and unfortunately stumbled upon such an error:

Can you share the full log? I also have a more recent version of the installer here https://github.com/proppy/conda-eda/releases/tag/v0.0-1418-g170b751

> What bothers me a little is that we'll have a tag added for each run of the Construct workflow, i.e., with each Upload workflow run from master. Do I get it right? Isn't there any other option to publish the installer?

I couldn't think of something else if we want to publish the installer as a GitHub release, another possibility it to move them on GCS in timestamped directory, that's something I think would also make sense for packages (see https://github.com/hdl/conda-eda/issues/204).

ajelinski:
Log: [openlane.log](https://github.com/hdl/conda-eda/files/9811071/openlane.log)

.github/workflows/Construct.yml
proppy: I think that would be `${{ env.GITHUB_WORKSPACE }}/openlane-sky130a-0-Linux-x86_64.sh`

umarcor: Using `openlane-sky130a-0-Linux-x86_64.sh` should also work. Since `actions/checkout@v3` is used, the default location for other steps is the root of the repo already.

.github/workflows/Construct.yml
```diff
@@ -0,0 +1,49 @@
+name: Build installers
```
umarcor, line 1: Why is this a separated workflow instead of an additional job in Build?

installer/openlane.sky130a/construct.yaml
```diff
@@ -0,0 +1,15 @@
+name: openlane-sky130a
+version: 0
+
+channels:
+  - https://conda.anaconda.org/litex-hub
```
umarcor, line 5: `litex-hub` alone should also work, isn't it?

proppy: Since all the constructor example were using fully qualified URLs: https://github.com/conda/constructor/search?q=conda.anaconda.org I assumed it was best-practice.

.github/workflows/Construct.yml
umarcor: Not the same. Construct is a new workflow, which is called as a reusable workflow from Upload. I mean having it as an additional job in workflow Build, without creating a new (reusable) workflow.
The Build workflow might be executed directly, not as a reusable workflow, which would ignore the recipes/jobs defined in Construct. However, if you put this job in Build, then it will always be run together with other recipes.

installer/openlane.sky130a/construct.yaml
ajelinski: It seems channel names are enough for other channels than main indeed: https://github.com/restlessronin/lila-deep/blob/415fc7172c49e0a1ae766c7c9331d0d85645ef33/.colab/construct.yaml

Though I found no `construct.yaml` with plain `main` on GH (https://github.com/search?q=main+extension%3Ayaml+filename%3Aconstruct.yaml&type=Code) so I have no preference on full URLs vs channel names.

.github/workflows/Construct.yml
ajelinski: FWIW I have no preference on the two.

.github/workflows/Construct.yml
```diff
@@ -0,0 +1,49 @@
...
+          path: |
+            /home/runner/work/conda-eda/conda-eda/openlane-sky130a-0-Linux-x86_64.sh
+
+      - run: |
+          git config --local user.name conda-eda@bot
+          git config --local user.email conda-eda@bot
+          TAG=$(git describe --tags)
+          git tag $TAG
+          git push origin "HEAD:refs/tags/$TAG"
+          echo "TAG=$TAG" >> $GITHUB_ENV
```
ajelinski, line 42: I've just realized there are two issues here:
1. `git describe --tags` will use tags from previous runs of this code in `Construct.yml` so pretty soon we'll have tags like: `v0.0-1337-gc24a15291-X-gSHA-X-gSHA-X-gSHA-X-gSHA-...`
2. There's no check if the tag already exists and the workflow will be triggered every day (with `Upload.yml` nightly runs).

.github/workflows/Construct.yml
ajelinski: > Not the same. Construct is a new workflow, which is called as a reusable workflow from Upload. I mean having it as an additional job in workflow Build, without creating a new (reusable) workflow. The Build workflow might be executed directly, not as a reusable workflow, which would ignore the recipes/jobs defined in Construct. However, if you put this job in Build, then it will always be run together with other recipes.

Good point that it shouldn't be in fact run from `Upload` but from `Build`. Perhaps you could move the Construct job then @proppy and use `needs: "openlane-linux"` for it? It'd be also nice to have `openlane.sky130a` parametrized in Construct.yml so that it's reusable if we ever want to have more installers constructed.

.github/workflows/Construct.yml
proppy: > 1. 

Oh I did had this issue, and handled it by removing `--tags` from the `git describe` command

> 2.

You're right it will overwrite the existing release with a new build even if there is no change in `conda-eda` which could be unfortunate if there are some upstream change but no `conda-eda` commit.

.github/workflows/Construct.yml
proppy: > The Build workflow might be executed directly, not as a reusable workflow, which would ignore the recipes/jobs defined in Construct. However, if you put this job in Build, then it will always be run together with other recipes.

My thinking that whe only want to "release" the installer when we actually publish the packages (so w/ that logic the same way we don't run `Upload` when we run `Build` independently, we couldn't run `Construct`).

proppy:
@PiotrZierhoffer, @umarcor would it make sense to setup a dedicated bucket for those so that we can get rid of the noise that automated tagging would add to the repo (as @ajelinski pointed in https://github.com/hdl/conda-eda/pull/247#pullrequestreview-1145791901)

PR #270, opened 2022-12-23
ci/conda-get: use bash to run miniconda

proppy:
Looks like miniconda started using bashism:
```
  ./Miniconda3-latest-Linux-x86_64.sh: 444: [[: not found
```

See https://github.com/google/CFU-Playground/issues/752


mithro (approved):
LGTM

PR #272, opened 2022-12-26
Disable interprocedural optimization for VTR

poname:
By default the `CMAKE_INTERPROCEDURAL_OPTIMIZATION` flag is on for VTR, this may cause `LTO` version incompatibility when using the binaries out of `VTR` scope in other projects.
Thir PR disables `INTERPROCEDURAL_OPTIMIZATION` for `VTR` package while installing by turning off the `VTR_IPO_BUILD` flag in the build script.

### DIFF
--- a/pnr/vtr-gui/build.sh
+++ b/pnr/vtr-gui/build.sh
@@ -13,7 +13,7 @@ case "${UNAME_OUT}" in
                 exit;;
 esac
 
-make V=1 CMAKE_PARAMS="-DCMAKE_INSTALL_PREFIX=${PREFIX}" -j$CPU_COUNT
+make V=1 CMAKE_PARAMS="-DCMAKE_INSTALL_PREFIX=${PREFIX} -DVTR_IPO_BUILD=off" -j$CPU_COUNT
 if [[ $OS != "Mac" ]]; then
 	make test
 fi


poname:
@mithro the binaries from VTR package are needed in [yosys-f4pga-plugins](https://github.com/chipsalliance/yosys-f4pga-plugins). LTO is enabled by default and creates issues when linking with other projects.

mithro:
Do you have more information about the LTO issues?

poname:
As an example when linking yosys-f4pga-plugins with the binaries from VTR package (litex-hub) the following error pops out:
- build locally
```
lto1: fatal error: bytecode stream in file ‘/lib/libarchfpga.a’ generated with LTO version 11.0 instead of the expected 11.3
compilation terminated.
lto-wrapper: fatal error: g++ returned 1 exit status
compilation terminated.
/usr/bin/ld: error: lto-wrapper failed
collect2: error: ld returned 1 exit status
```
- when running the CI:
```
lto1: fatal error: bytecode stream in file ‘/home/runner/work/yosys-f4pga-plugins/yosys-f4pga-plugins/env/conda/envs/yosys-plugins/lib/libarchfpga.a’ generated with LTO version 12979.5 instead of the expected 8.1
  compilation terminated.
  lto-wrapper: fatal error: /usr/bin/g++ returned 1 exit status
  make[1]: Leaving directory '/home/runner/work/yosys-f4pga-plugins/yosys-f4pga-plugins/parmys-plugin'
  compilation terminated.
  /usr/bin/ld: error: lto-wrapper failed
  collect2: error: ld returned 1 exit status
```
Perhaps the problem is from the compiler versions. I think the best solution is to disable LTO for VTR package and according to the VTR documentation, it only improves the performance and disabling should cause no harm.

poname:
Also, when the VTR binaries are built on the same machine that the linking is being performed later, there are no issues since the compilers are of the exact same version (and the system architectures are the same) . I'm no expert but AFAIK LTO depends on the compiler version and system architecture.

poname:
@mithro Do we need any changes?

mithro:
@poname - I think the error points out there is something going wrong with the compiler toolchain being used. The fix is to understand that problem rather than disabling LTO.

poname:
@mithro I am usin the default compiler toolchain in [yosys-f4pga-plugins](https://github.com/chipsalliance/yosys-f4pga-plugins) build and I think resolvig the LTO version issue requires changing the compiler there, which is unlikely to be accepted.

Instead of changing the main `VTR` package here, can we have a new package named `VTR-libs` but whithout LTO?

mithro:
@poname - I'm pretty sure the `yosys-f4pga-plugins` repository **_should_** be using the same compiler as `hdl/conda-eda` when building against tools from the `hdl/conda-eda` repository. The fact that it is not is somewhat concerning.

poname:
from the `vtr-linux` ci log in `hdl/conda-eda` actions:
`  -- The CXX compiler identification is GNU 11.2.0
`

while the `chipsalliance/yosys-f4pga-plugins` explicitly uses `g++-9`:
`        sudo apt-get install git g++-9 build-essential bison flex \ ...
`

I am closing this PR to keep other packages intact and a new PR will be added to introduce a new package named [`vtr-libs`](https://github.com/hdl/conda-eda/pull/275) to provide a lightweight package of only VTR binaries with their header files.

mithro:
@poname - If these are building against `hdl/conda-eda` then they should be also be using the compilers from conda not from the local system.

poname:
@mithro Yes,the `chipsalliance/yosys-f4pga-plugins` ci is not using the conda c/cxx compilers. This is hardcoded within the [ci scripts](https://github.com/chipsalliance/yosys-f4pga-plugins/blob/ca5bb9a2bab9e264d70eab71b1067ad27713afb2/.github/workflows/ci.yml#L34-L40)  and I think changing their global ci ecosystem is not an option.

Since playing with build flags would solve the problem, if it is OK from your side we can have a new mini package of vtr binaries and use it without touching anything else.

mithro:
@poname - I think we need to chat with `chipsalliance/yosys-f4pga-plugins` and get them to fix that.

mithro:
@poname - Can you point out where you think they are not using the conda compilers?

poname:
@mithro check here: https://github.com/chipsalliance/yosys-f4pga-plugins/blob/main/.github/workflows/ci.yml

they are using `apt-get install` to get dependencies.

```
    - name: Install
      run: |
        sudo apt-get update
        sudo apt-get install git g++-9 build-essential bison flex \
          libreadline-dev gawk tcl-dev libffi-dev git graphviz xdot \
          pkg-config libboost-system-dev libboost-python-dev \
          libboost-filesystem-dev zlib1g-dev clang-format-8 cmake
```

PR #273, opened 2023-01-03
syn/yosys: Update failing patch and run tests in the test stage instead of during building

ajelinski:


ajelinski:
None of the failures is caused by these changes in the Yosys package.
The plugins fail because of surelog (see https://github.com/chipsalliance/yosys-f4pga-plugins/pull/428 and https://github.com/chipsalliance/yosys-f4pga-plugins/issues/429) and the OpenLane failure is due to a newer magic package which cannot be used by the older open_pdks and that building open_pdks fail at the same time.

The OpenLane was tested locally to work well with the Yosys package built by this CI so I'm merging.

PR #274, opened 2023-01-03
Add tbb to run dependencies for nextpnr-fpga-interchange

rw1nkler:
The tbb library, besides being required during build time, is also required to be in a specific version when using the nextpnr-fpga-interchange package.


PiotrZierhoffer (commented):
LGTM, mergable if it works

PR #275, opened 2023-01-04
[WIP] vtr-libs

poname:


misc/vtr-libs/meta.yaml
```diff
@@ -0,0 +1,49 @@
...
+    - bison 3.4
+    - cmake
+    - flex
+    - pkg-config
+    - tbb <2021.0.0a0
+    - tbb-devel <2021.0.0a0
+  run:
+    - tbb <2021.0.0a0
+
+#test:
```
proppy, line 40: curious why we disable the tests?

misc/vtr-libs/meta.yaml
```diff
@@ -0,0 +1,49 @@
+# Use `conda-build-prepare` before building for a better version string.
```
proppy, line 1: shouldn't this go alongside the other `vtr` packages in https://github.com/hdl/conda-eda/tree/master/pnr?

misc/vtr-libs/build.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
```
proppy, line 1: this looks very similar to https://github.com/hdl/conda-eda/blob/master/pnr/vtr-gui/build.sh, I noticed https://github.com/hdl/conda-eda/tree/master/pnr/vtr does a symlink to the vtr-gui package, maybe something similar could be done here?

misc/vtr-libs/meta.yaml
```diff
@@ -0,0 +1,49 @@
...
+  name: vtr-libs
+  version: {{ version }}
+
+source:
+  git_url: https://github.com/verilog-to-routing/vtr-verilog-to-routing.git
+  git_rev: master
+  patches:
+    - osx.patch             [osx]
+
+build:
```
proppy, line 14: I wonder if we could leverage conda outputs, to have a single build for vtr that ends up in multiple conda packages.

Maybe @ajelinski could advise on the best way to consolidate this?

misc/vtr-libs/meta.yaml
```diff
@@ -0,0 +1,49 @@
...
+#test:
+#  commands:
+#    - ./run_reg_test.pl vtr_reg_basic
+#    - ./run_reg_test.pl vtr_reg_strong -j2
+
+about:
+  home: http://verilogtorouting.org/
+  license: MIT
+  license_file: LICENSE.md
+  summary: 'The Verilog-to-Routing (VTR) project is a world-wide collaborative effort to provide a open-source framework for conducting FPGA architecture and CAD research and development. The VTR design flow takes as input a Verilog description of a digital circuit, and a description of the target FPGA architecture.'
```
proppy, line 49: maybe we should expand in https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#about-section about the differences with the other vtr packages?

PR #277, opened 2023-01-05
CI: Only build the OpenLane package if building Open-PDKs succeeded

ajelinski:
This change will ensure a problem when testing OpenLane in the same CI that:
* built a new version of the magic package,
* failed to build `open_pdks` with that new magic version
won't happen again since debugging this isn't trivial.

It turned out all of the open_pdks packages newer than 1.0.286 have a strict magic version requirement. The just-built magic package will be always preferred in CI so when building `open_pdks` fail and the CI built a new version of the magic package then this old `1.0.286` Open-PDKs package will be used which doesn't work well with the other latest packages.

Such a run happened in https://github.com/hdl/conda-eda/actions/runs/3828789250/jobs/6551196431#step:7:3305

### DIFF
--- a/.github/workflows/Build.yml
+++ b/.github/workflows/Build.yml
@@ -722,7 +722,7 @@ jobs:
 
   #64
   openlane-linux:
-    needs: ["magic-linux", "netgen-linux", "openroad-linux", "yosys-linux-py37", "yosys-linux-py38"]
+    needs: ["magic-linux", "netgen-linux", "open_pdks-linux", "openroad-linux", "yosys-linux-py37", "yosys-linux-py38"]
     runs-on: "ubuntu-20.04"
     env:
       PACKAGE: "misc/openlane"


proppy:
LGTM! Thanks for pinning this down. Maybe we could also delete version of open_pdks older than `1.0.286`? (That could reclaim precious disk space on the channel)

ajelinski:
> LGTM! Thanks for pinning this down. Maybe we could also delete version of open_pdks older than `1.0.286`? (That could reclaim precious disk space on the channel)

Are you sure these old packages aren't used by anyone?

Btw. perhaps we should add a constraint to the OpenLane package so that open_pdks has to be newer than that? It can be an optional one, i.e., not making any open_pdks required for running OpenLane. Also it'd be best to have this compatibility set more precisely since now the latest OpenLane allows using any openroad and open_pdks versions but unfortunately I don't know enough about these tools to establish it.

proppy:
> Btw. perhaps we should add a constraint to the OpenLane package so that open_pdks has to be newer than that? It can be an optional one, i.e., not making any open_pdks required for running OpenLane.

Yes, I was thinking about the same thing, maybe it could leverage https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#run-constrained, like the open_pdks package does with `magic`:
https://github.com/hdl/conda-eda/blob/master/misc/open_pdks/meta.yaml#L47-L48

PR #279, opened 2023-01-10
Unpin TBB dependency for Nextpnr FPGA Interchange

mkurc-ant:
Removed the requirement for TBB version for Nextpnr FPGA Interchange so that `nextpnr-fpga_interchange` won't conflict with `vtr-optimized`.


ajelinski (approved):
LGTM

PR #291, opened 2023-02-13
[DRAFT] custom runners cleanup openroad fix

PiotrZierhoffer:


proppy:
Fixed openroad with #294 and fixing openlane with #296

PR #293, opened 2023-02-15
Add support for yosys and iverilog on osx arm64

xobs:
This adds support for building these packages on arm64 versions of macos.

Note that the biggest change is the clang compiler -- only clang 12 and clang 14 are available in conda, so I picked the latest available version.

### DIFF
--- a/sim/icarus/add-arm64.patch
+++ b/sim/icarus/add-arm64.patch
@@ -1,0 +1,3118 @@
+diff --git a/config.sub b/config.sub
+index 8f1229c6f..d74fb6dea 100755
+--- a/config.sub
++++ b/config.sub
+@@ -1,8 +1,10 @@
+ #! /bin/sh
+ # Configuration validation subroutine script.
+-#   Copyright 1992-2015 Free Software Foundation, Inc.
++#   Copyright 1992-2021 Free Software Foundation, Inc.
+ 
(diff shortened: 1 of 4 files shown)


xobs:
cc @proppy

proppy:
@PiotrZierhoffer @ajelinski do we have some CI environment for osx/arm64?

PiotrZierhoffer:
@proppy @xobs only GitHub runners, vide https://github.com/hdl/conda-eda/blob/master/.github/workflows/Build.yml#L335-L342

proppy:
> @proppy @xobs only GitHub runners

But that's x64 right not arm64?

xobs:
Yes, that's x64 not arm64.

The magical incantation to switch to arm64 is `conda config --env --set subdir osx-arm64`, but I suspect that won't work on the github builders since they don't build fat binaries and you can't run arm64 binaries on x64.

sim/icarus/add-arm64.patch
```diff
@@ -0,0 +1,3118 @@
+diff --git a/config.sub b/config.sub
```
proppy, line 1: should we copy that from the autoconf package like you did for #303 ?

xobs: Good idea. I've copied it over.

PR #295, opened 2023-02-15
misc/open_pdks: fix source handling

proppy:
https://github.com/hdl/conda-eda/commit/a852977eea1fd6b3879eb7cab9567983c074e4ac changed meta.yaml to use `git_url` causing conda to fetch gitsubmodules recursively.

Reverting back to tarball allow us to make sure we only clone each dependencies once.


mithro (approved):
I think there must be a better solution (please create a github issue), but for now let's just get things working again.

proppy:
Seems like we're hitting https://github.com/conda/conda-build/issues/4775, so we should make sure we point to archive that are named differently.

PR #297, opened 2023-02-17
pnr/openroad: add osx build

xobs:
This simple patch gets openroad building on arm64:

```
sean in 🌐 Garlemald in conda-eda on  openroad-osx-arm64 [$?] via 🅒 eda  took 17s
❯ openroad -version                                                                                                                                            (eda)
5c85b36f8d50975d4e57af6678a0bbfe62046f0e

sean in 🌐 Garlemald in conda-eda on  openroad-osx-arm64 [$?] via 🅒 eda
❯ uname -a                                                                                                                                                     (eda)
Darwin Garlemald.local 22.3.0 Darwin Kernel Version 22.3.0: Mon Jan 30 20:39:35 PST 2023; root:xnu-8792.81.3~2/RELEASE_ARM64_T8103 arm64

sean in 🌐 Garlemald in conda-eda on  openroad-osx-arm64 [$?] via 🅒 eda
❯                                                                                                                                                              (eda) 
```


xobs:
cc @proppy

pnr/openroad/meta.yaml
```diff
@@ -41,16 +41,18 @@ requirements:
     - python
     - libboost 1.73
     - eigen 3.3
-    - spdlog 1.8.1
+    - spdlog 1.9
```
proppy, line 44: >=1.9.2?

xobs: Is 1.9.2 available on all platforms? I thought I could just specify 1.9 and it would pick any version that was available on that platform. If 1.9.2 is universally available, I'll switch it to that.

pnr/openroad/meta.yaml
```diff
@@ -41,16 +41,18 @@ requirements:
...
     - zlib
     - fmt
-    - libgomp
+    - libgomp                 [not osx]
+    - llvm-openmp             [osx]
   run:
     # missing run_exports
     - {{ pin_compatible('libboost', min_pin='x.x', max_pin='x.x') }}
     - {{ pin_compatible('python', min_pin='x.x', max_pin='x.x') }}    
+    - spdlog >=1.9,<1.10
```
proppy, line 55: This is normally not necessary because package have `run_exports` directive, but here it seems that this one is looser and will allow `1.x`:
https://github.com/AnacondaRecipes/spdlog-feedstock/blob/master/recipe/meta.yaml#L31

xobs:
Note that there is a separate entry under `run` for spdlog. Before that line was added, the package built with spdlog 1.9.2 but installed with 1.11.0, which is incompatible.

pnr/openroad/meta.yaml
proppy: Looks like 1.9.0 and 1.9.2 are available: so 1.9 sounds good then!

xobs:
The relevant issue on spdlog: https://github.com/The-OpenROAD-Project/OpenROAD/issues/2220

pnr/openroad/meta.yaml
proppy: https://anaconda.org/main/spdlog/files

proppy (approved):

xobs:
Sure. I'm guessing these runners don't actually have the ability to execute software, which is why things with tests have `--no-test`? Should I also skip tests on openroad?

proppy:
> Sure. I'm guessing these runners don't actually have the ability to execute software, which is why things with tests have --no-test? Should I also skip tests on openroad?

I think they do, but they would build the x64 version (not arm64).

xobs:
Alright, I think that adds a `Build` step that works.

proppy:
Some build error on macosx x64:
```
2023-02-17T14:06:23.9687350Z /Users/runner/work/conda-eda/conda-eda/workdir/conda-env/conda-bld/openroad_1676639035640/work/src/drt/src/dr/FlexDR_graphics.cpp:69:20: error: 'any_cast<fr::GridGraphDescriptor::Data>' is unavailable: introduced in macOS 10.13 - see https://conda-forge.org/docs/maintainer/knowledge_base.html#newer-c-features-with-old-sdk
2023-02-17T14:06:23.9789260Z   auto data = std::any_cast<Data>(object);
2023-02-17T14:06:23.9794460Z                    ^
2023-02-17T14:06:23.9896180Z /Users/runner/work/conda-eda/conda-eda/workdir/conda-env/conda-bld/openroad_1676639035640/_build_env/bin/../include/c++/v1/any:611:12: note: 'any_cast<fr::GridGraphDescriptor::Data>' has been explicitly marked unavailable here
```

proppy:
https://conda-forge.org/docs/maintainer/knowledge_base.html#newer-c-features-with-old-sdk seems to imply that this is safe to disable with `CXXFLAGS="${CXXFLAGS} -D_LIBCPP_DISABLE_AVAILABILITY"` but I wonder if it also apply to non-conda forge c++ toolchain (I guess yes, since it mentions that url in the error message).

for the record github runner environment include the following MacOSX SDKs: https://github.com/actions/runner-images/blob/main/images/macos/macos-12-Readme.md#installed-sdks

xobs:
@proppy I believe I added `D_LIBCPP_DISABLE_AVAILABILITY` to the build. Would you mind trying the build again?

xobs:
Looks like openroad-osx built this time

Issue #72, opened 2021-01-27
Build all packages requiring Python with versions: 3.7, 3.8 and 3.9


ajelinski:
@mithro (https://github.com/litex-hub/litex-conda-eda/pull/68#issuecomment-767706800):
> All packages which require Python should build for Python 3.7, Python 3.8, Python 3.9
> 
> The reason I was suggesting everything is there is some subtle complexity around the dependencies of Python itself can conflict with dependencies of the other packages.

This can be done in the same way as in the `surelog`, `yosys` and `symbiyosys` recipes from #68.

mithro:
@ajelinski - We should figure out how to do this automatically? We should always be supporting the latest 3 versions of Python.

PR #298, opened 2023-02-18
Fix klayout build on macos

xobs:
This fixes the klayout package so that it now builds properly on `osx-arm64`.

### DIFF
--- a/.github/workflows/Build.yml
+++ b/.github/workflows/Build.yml
@@ -341,6 +341,15 @@ jobs:
       - uses: actions/checkout@v3
       - uses: ./ci
 
+  klayout-osx:
+    runs-on: "macos-latest"
+    env:
+      PACKAGE: "misc/klayout"
+      OS_NAME: "osx"
+    steps:
+      - uses: actions/checkout@v3
+      - uses: ./ci
+
   magic-osx:
     runs-on: "macos-latest"
     env:
--- a/misc/klayout/build.sh
+++ b/misc/klayout/build.sh
@@ -7,10 +7,29 @@ set -x
 UNAME_OUT="$(uname -s)"
 case "${UNAME_OUT}" in
     Linux*)     OS=Linux;;
+    Darwin*)    OS=Mac;;
     *)          OS="${UNAME_OUT}"
                 echo "Unknown OS: ${OS}"
                 exit;;
 esac
+
+if [[ $OS == "Linux" ]]; then
+    bin_ext=""
+    lib_ext=".so"
+elif [[ $OS == "Mac" ]]; then
+    export bin_ext=".app"
+    export lib_ext=".dylib"
+    cd ${PREFIX}
+    if grep -q -- '-isysroot $$sysroot_path $$version_min_flag' mkspecs/features/mac/default_post.prf; then
+        sed 's|-isysroot $$sysroot_path $$version_min_flag|-isysroot '$CONDA_BUILD_SYSROOT' -mmacosx-version-min='$MACOSX_DEPLOYMENT_TARGET'|g' mkspecs/features/mac/default_post.prf > mkspecs/features/mac/default_post.prf.bkp
+        mv mkspecs/features/mac/default_post.prf.bkp mkspecs/features/mac/default_post.prf
+    elif grep -q -- '-isysroot $$QMAKE_MAC_SDK_PATH $$version_min_flag' mkspecs/features/mac/default_post.prf; then
+        sed 's|-isysroot $$QMAKE_MAC_SDK_PATH $$version_min_flag|-isysroot '$CONDA_BUILD_SYSROOT' -mmacosx-version-min='$MACOSX_DEPLOYMENT_TARGET'|g' mkspecs/features/mac/default_post.prf > mkspecs/features/mac/default_post.prf.bkp
+        mv mkspecs/features/mac/default_post.prf.bkp mkspecs/features/mac/default_post.prf
+    fi
+    sed 's|^QMAKE_MAC_SDK_PATH =.*|QMAKE_MAC_SDK_PATH = "'$CONDA_BUILD_SYSROOT'"|g' mkspecs/features/mac/sdk.prf > mkspecs/features/mac/sdk.prf.bkp
+    mv mkspecs/features/mac/sdk.prf.bkp mkspecs/features/mac/sdk.prf
+fi
 
 cd ${SRC_DIR}
 ./build.sh -build "${SRC_DIR}/build" -python "${PYTHON}" -expert -without-qtbinding -libpng -libexpat -dry-run
@@ -20,5 +39,13 @@ make V=1 -j$CPU_COUNT
 make V=1 install
 
 cd ${SRC_DIR}/bin-release
-cp -a klayout strm* ${PREFIX}/bin/
-cp -ar *.so* pymod *_plugins ${PREFIX}/lib/
+cp -a klayout${bin_ext} strm* ${PREFIX}/bin/
+cp -a *${lib_ext}* pymod *_plugins ${PREFIX}/lib/
+
+if [[ $OS == "Mac" ]]; then
+    # Add a symlink to allow it to run from the command line
+    cd ${PREFIX}/bin/
+    ln -s klayout${bin_ext}/Contents/MacOS/klayout .
+    cd ${SRC_DIR}
+    cp -a build/pymod/*${lib_ext}* ${PREFIX}/lib/pymod/
+fi
--- a/misc/klayout/meta.yaml
+++ b/misc/klayout/meta.yaml
@@ -27,32 +27,36 @@ build:
 requirements:
   build:
     - make
+    - {{ compiler('c') }}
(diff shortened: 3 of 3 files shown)


xobs:
Note that I'm not sure why `build/pymod/*.dylib` isn't copied over to `bin-release/pymod/`, but it seems to be copied on the Linux version. So I added the final cp for only that platform.

proppy:
> Note that I'm not sure why build/pymod/*.dylib isn't copied over to bin-release/pymod/, but it seems to be copied on the Linux version. So I added the final cp for only that platform.

Maybe @klayoutmatthias as a clue?

misc/klayout/build.sh
```diff
@@ -7,18 +7,34 @@ set -x
...
 cd ${SRC_DIR}
-./build.sh -build "${SRC_DIR}/build" -python "${PYTHON}" -expert -without-qtbinding -libpng -libexpat -dry-run
+if [[ $OS == "Linux" ]]; then
+    ./build.sh -build "${SRC_DIR}/build" -python "${PYTHON}" -expert -without-qtbinding -libpng -libexpat -dry-run
+    bin_ext=""
+    lib_ext=".so"
+elif [[ $OS == "Mac" ]]; then
+    # Support `any_cast` workaround: https://conda-forge.org/docs/maintainer/knowledge_base.html#newer-c-features-with-old-sdk
+    export CXXFLAGS="${CXXFLAGS} -D_LIBCPP_DISABLE_AVAILABILITY"
+    ./build.sh -build "${SRC_DIR}/build" -python "${PYTHON}" -expert -without-qtbinding -libpng -libexpat -dry-run
```
proppy, line 24: Maybe we can move this out of the if because it's the same on both platform?

misc/klayout/build.sh
```diff
@@ -7,18 +7,34 @@ set -x
...
 make V=1 install
 
 cd ${SRC_DIR}/bin-release
-cp -a klayout strm* ${PREFIX}/bin/
-cp -ar *.so* pymod *_plugins ${PREFIX}/lib/
+cp -a klayout${bin_ext} strm* ${PREFIX}/bin/
+cp -a *${lib_ext}* pymod *_plugins ${PREFIX}/lib/
+
+if [[ $OS == "Mac" ]]; then
+    cd ${SRC_DIR}
```
proppy, line 38: Maybe we could patch the Makefile to do this?

xobs: It seems like it should do this as part of the install, but it doesn't, and I'm not sure why not.

misc/klayout/build.sh
xobs: I thought these commands might need to be different, but surprisingly the same command actually works!

I'll combine the two commands.

xobs:
I added a symlink to `klayout`, which allows it to be used from the command line.

.github/workflows/Build.yml
```diff
@@ -341,16 +341,22 @@ jobs:
...
     steps:
       - uses: actions/checkout@v3
       - uses: ./ci
 
+  openroad-osx:
+    runs-on: "macos-latest"
+    env:
+      PACKAGE: "pnr/openroad"
+      OS_NAME: "osx"
+      EXTRA_BUILD_ARGS: "--no-test"
```
proppy, line 358: missing steps?

xobs: Good catch. That was a weird merge. Fixed.

proppy:
Looks like the osx build failed with:
```
2023-02-18T11:44:41.6904360Z ld: warning: -pie being ignored. It is only used when linking a main executable
2023-02-18T11:44:41.7225230Z ld: in '/Applications/Xcode_14.2.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX13.1.sdk/System/Library/Frameworks//ColorSync.framework/Versions/A/ColorSync.tbd', unsupported tapi file type '!tapi-tbd' in YAML for architecture x86_64
2023-02-18T11:44:41.7241560Z clang-14: error: linker command failed with exit code 1 (use -v to see invocation)
2023-02-18T11:44:41.7265570Z make[2]: *** [Makefile:384: ../../tl/libtl_tests.0.28.5.dylib] Error 1
2023-02-18T11:44:41.7284780Z make[1]: *** [Makefile:76: sub-unit_tests-make_first] Error 2
2023-02-18T11:44:41.7296980Z make: *** [Makefile:74: sub-tl-make_first] Error 2
```

xobs:
Conda seems to be adding `-isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk -mmacosx-version-min=10.15` on my local build, and I can't figure out where that's coming from. 

What's even more annoying is that the environment doesn't seem to have those variables anywhere. If I modify `build.sh` to use the correct values, it appears twice:

```
    export CXXFLAGS="${CXXFLAGS} -mmacosx-version-min=${MACOSX_DEPLOYMENT_TARGET} -D_LIBCPP_DISABLE_AVAILABILITY -isysroot ${CONDA_BUILD_SYSROOT}"
    export CFLAGS="${CFLAGS} -mmacosx-version-min=${MACOSX_DEPLOYMENT_TARGET} -D_LIBCPP_DISABLE_AVAILABILITY -isysroot ${CONDA_BUILD_SYSROOT}"
```

becomes:

```
...
-mmacosx-version-min=11.0 \
-D_LIBCPP_DISABLE_AVAILABILITY \
-isysroot /Users/sean/Code/EDA/sdk/MacOSX10.9.sdk \
-std=c++17 \
-fvisibility=hidden \
-std=c++17 \
-fvisibility=hidden \
-std=gnu++1z \
$(EXPORT_ARCH_ARGS) \
-isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk \
-mmacosx-version-min=10.15
...
```

xobs:
...also, I'm not sure where it's getting `MACOSX_DEPLOYMENT_TARGET=10.15` *or* `MACOSX_DEPLOYMENT_TARGET=11.0` from. Both `ci/conda_build.config.yaml` and the build environment set `MACOSX_DEPLOYMENT_TARGET=10.9`, and that's what Conda seems to imply is the default:

https://docs.conda.io/projects/conda-build/en/latest/user-guide/environment-variables.html#env-vars

klayoutmatthias:
> > Note that I'm not sure why build/pymod/*.dylib isn't copied over to bin-release/pymod/, but it seems to be copied on the Linux version. So I added the final cp for only that platform.
> 
> Maybe @klayoutmatthias as a clue?

Honestly, I have never tested pymod generation on MacOS. Mainly because MacOS is not my home turf. The installation basically is generic, but maybe it needs to be specialized for MacOS.

xobs:
I discovered that Qt ignores standard CFLAGS and CXXFLAGS and instead uses its own QMAKE_CFLAGS and QMAKE_CXXFLAGS.

I also discovered that MacOS 11.0 was the first version that supported arm64, so I bumped the minimum version up to that.

Note that 10.11 is the first version that supported Metal, so we would need to use that at a minimum in order to get support for klayout.

Additionally, it looks like it might be possible to build osx-arm64 from osx-x86_64 as documented at https://conda-forge.org/blog/posts/2020-10-29-macos-arm64/ -- note that they also build for macos 11.0 in there.

proppy:
> Additionally, it looks like it might be possible to build osx-arm64 from osx-x86_64 as documented at https://conda-forge.org/blog/posts/2020-10-29-macos-arm64/ -- note that they also build for macos 11.0 in there.

Filed https://github.com/hdl/conda-eda/issues/299 to discuss arm64 builds.

proppy (approved):

xobs:
It builds and passes tests! \o/

.github/workflows/Build.yml
```diff
@@ -341,6 +341,15 @@ jobs:
       - uses: actions/checkout@v3
       - uses: ./ci
 
+  klayout-osx:
+    runs-on: "macos-latest"
+    env:
+      PACKAGE: "misc/klayout"
+      OS_NAME: "osx"
```
proppy, line 348: I assume we deliberatly don't set `PYTHON_VERSION` here like the linux job does it.

xobs: What purpose would that serve? Wouldn't it make more sense to use the same Python version as the rest of the environment?

proppy: I think the general idea of how `PYTHON_VERSION` is set in Build.yaml is to explicit state upfront the version of the package we want to tie the package again so that we don't end up being dependent of the builder environment (and we could add additional jobs as part of #72).

For osx we could use a different python version but still be explicit about it?

xobs: Alright, I've added explicit packages for python-3.8 and python-3.10.

proppy:
Looks like `python3.10` job is failing when trying to install `python >= 3.10` as a host dep from conda running on `python 3.7`.

xobs:
I'll just remove that, then...

proppy:
Merging since both jobs passes (not sure if we want to suffix the job with `-python38` for consistency).

PR #301, opened 2023-02-24
misc/openlane: set openlane to be noarch

xobs:
openlane is tcl scripts, so there is no reason for it to be arch-specific.


proppy:
That make sense, with the caveat that some of the deps might not yet resolve on all archs.

proppy:
We might want to disable `RUN_LVS` in the default conf until we get netgen packaged for OSX.

xobs:
Additional changes needed to be made to build.sh in order to get it building on something other than Linux.

proppy (approved):

proppy:
merging since both netgen and tcllib are in!

PR #302, opened 2023-02-24
Build magic on osx

xobs:
Fix magic so that it can build on osx, albeit without X11.

This requires a patch, and is waiting on https://github.com/RTimothyEdwards/magic/pull/222, at which point we can remove the patch.


proppy (approved):

proppy:
Thanks for working on this!

PR #303, opened 2023-02-24
Build netgen on macos

xobs:
Just disable X11, since Conda doesn't include the requisite packages.

### DIFF
--- a/misc/netgen/build.sh
+++ b/misc/netgen/build.sh
@@ -7,6 +7,7 @@ set -x
 UNAME_OUT="$(uname -s)"
 case "${UNAME_OUT}" in
     Linux*)     OS=Linux;;
+    Darwin*)    OS=Mac;;
     *)          OS="${UNAME_OUT}"
                 echo "Unknown OS: ${OS}"
                 exit;;
@@ -22,12 +23,23 @@ echo "---------------------"
 
 echo CPPFLAGS="$CPPFLAGS"
 
-./configure \
-	--prefix="${PREFIX}" \
-	--with-tk="${PREFIX}" \
-	--with-tcl="${PREFIX}" \
-	--x-includes="${PREFIX}" \
-	--x-libraries="${PREFIX}"
+# Get an updated config.sub and config.guess
+cp $BUILD_PREFIX/share/autoconf/build-aux/config.* scripts
+
+if [ "$OS" = "Linux" ]; then
+	./configure \
+		--prefix="${PREFIX}" \
+		--with-tk="${PREFIX}" \
+		--with-tcl="${PREFIX}" \
+		--x-includes="${PREFIX}" \
+		--x-libraries="${PREFIX}"
+elif [ "$OS" = "Mac" ]; then
+	./configure \
+		--prefix="${PREFIX}" \
+		--with-tk="${PREFIX}" \
+		--with-tcl="${PREFIX}" \
+		--without-x
+fi
 
 echo "Configure ended with: $?"
 make V=1 -j$NUM_CORES || exit 1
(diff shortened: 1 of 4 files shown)


misc/netgen/build.sh
```diff
@@ -22,12 +23,23 @@ echo "---------------------"
...
 echo CPPFLAGS="$CPPFLAGS"
 
-./configure \
-	--prefix="${PREFIX}" \
-	--with-tk="${PREFIX}" \
-	--with-tcl="${PREFIX}" \
-	--x-includes="${PREFIX}" \
-	--x-libraries="${PREFIX}"
+# Get an updated config.sub and config.guess
+cp $BUILD_PREFIX/share/autoconf/build-aux/config.* scripts
```
proppy, line 27: curious why this was necessary? Should we get those updated upstream?

proppy (approved):

misc/netgen/meta.yaml
```diff
@@ -31,6 +31,10 @@ requirements:
   run:
     - python
 
+test:
+  commands:
+    - netgen Q
```
proppy, line 36: Looks like the test assume a `DISPLAY`, can we use `netgen -batch quit` instead?

xobs: It works fine on my system. I think it only consults `DISPLAY` if compiled with X11.

proppy: Yep, but the Linux build doesn't currently disable X11 it seems. So this makes the CI fail.

xobs: Alright, let's see if that fixes it

misc/netgen/build.sh
xobs: Admittedly, I'm trying this on arm64, and the build is old enough that it doesn't even configure on arm64.

I adapted it from https://conda-forge.org/blog/posts/2020-10-29-macos-arm64/#how-to-add-a-osx-arm64-build-to-a-feedstock which didn't work on the conda version of gnuconfig.

This should probably be fixed upstream as well.

misc/netgen/remove-x11-requirement.patch
```diff
@@ -0,0 +1,38 @@
+diff --git a/scripts/configure b/scripts/configure
```
proppy, line 1: We should send this upstream to fix https://github.com/RTimothyEdwards/netgen/issues/9 at some point :)

xobs: I'm waiting to see if it actually builds.

It's entirely possible that it won't actually build, and that it's relying on include files that are leaking in from my environment. However, I can say for sure that it's building on my system with this patch, and without this patch it errored with trying to find libX11.

We'll need to see what CI says.

misc/netgen/remove-x11-requirement.patch
```diff
@@ -0,0 +1,38 @@
...
++     gr_dflags="$gr_dflags"
++     gr_libs="$gr_libs"
+      gr_srcs="$gr_srcs \${TK_SRCS}"
+   else
+      gr_srcs="$gr_srcs \${NULL_SRCS}"
+@@ -6094,7 +6094,6 @@ if test $usingTcl ; then
+ else
+   if test $usingX11 ; then
+     gr_dflags="$gr_dflags -DX11 -DXLIB"
+-    gr_libs="$gr_libs -lX11"
```
proppy, line 35: curious why you need to remove this line?

xobs: I admit that I wasn't sure which path it would take. I build using `--without-x`, yet it seems that it sometimes takes the `$usingX11` path instead.

In order to make sure either path results in the same thing, I applied the change to both branches.

PR #305, opened 2023-02-25
misc/openlane: patch sed to be cross-platform

xobs:
sed on BSD and GNU have different behaviour when it comes to the -i flag. They behave the same when a backup file is specified, but differently when no backup is specified.

Rework all invocations in order to get this working on osx.

There is a pending upstream patch: https://github.com/The-OpenROAD-Project/OpenLane/pull/1668


misc/openlane/build.sh
```diff
@@ -20,14 +20,24 @@ set -ex
 UNAME_OUT="$(uname -s)"
 case "${UNAME_OUT}" in
     Linux*)     OS=Linux;;
+<<<<<<< Updated upstream
```
proppy, line 23: did you meant to commit this?

xobs: I did not! I didn't realize that the upstream changed to use `cp` rather than `git`. This makes the patch a bit easier, thanks.

proppy:
@xobs sorry for dropping this, do you still want to get that in?

xobs:
In the intervening time, the openlane patches were merged upstream: https://github.com/The-OpenROAD-Project/OpenLane/commit/45980c85ecf4094c977785432a8ebb0adf0e562d

So we can just drop dbd35f2d7b589aac3201ef0a66732b16805264af from the patchset.

PR #307, opened 2023-03-07
meta: add downstream tests

### LINKED ISSUE #306: run downstream tests for reverse dependencies
Recently a commit from `open_pdks` https://github.com/RTimothyEdwards/open_pdks/commit/1683283a55ace46a24193f37b39cf271a5ab1808 broke one of the backward dependencies: https://anaconda.org/LiteX-Hub/openlane, causing some of the notebooks at https://github.com/chipsalliance/silicon-notebooks/issues/59 to be broken.

We could leverage `Downstream tests` https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#downstream-tests to ensure that we don't publish a new package that breaks its reverse dependencies.

proppy:
Fixes #306

### DIFF
--- a/misc/klayout/meta.yaml
+++ b/misc/klayout/meta.yaml
@@ -57,6 +57,8 @@ test:
   requires:
     - ruby
     - python {{ python_version }}
+  downstreams:
+    - openlane
 
 about:
   home: https://www.klayout.de/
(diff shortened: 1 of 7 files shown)


proppy:
@mithro fyi
@xobs am I missing any other package relation? ideally, we'd also like open_pdks to trigger test for all packages it integrate with: `xschem`, `klayout`, `magic`, `netgen`, `ngspice`, but we don't have yet tests that exercise the pdk integration for all of those.

PR #311, opened 2023-03-10
misc/open_pdks.gf180mcu: add pv repo

proppy:
/cc @xobs @mabrains fyi

### DIFF
--- a/misc/open_pdks/gf180mcuc/use-sources-from-conda.patch
+++ b/misc/open_pdks/gf180mcuc/use-sources-from-conda.patch
@@ -1,18 +1,20 @@
 diff --git a/gf180mcu/Makefile.in b/gf180mcu/Makefile.in
-index a40e9e0..98f3f2d 100644
+index da16966..2967c43 100644
 --- a/gf180mcu/Makefile.in
 +++ b/gf180mcu/Makefile.in
-@@ -151,12 +151,12 @@ endif
+@@ -152,13 +152,13 @@ endif
  
  # Path to GF180MCU open PDK sources.  If this is specified, then all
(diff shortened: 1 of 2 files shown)


misc/open_pdks/gf180mcuc/use-sources-from-conda.patch
```diff
@@ -1,18 +1,20 @@
...
+@@ -152,13 +152,13 @@ endif
  
  # Path to GF180MCU open PDK sources.  If this is specified, then all
  # library repositories must be subdirectories of this path.
 -GF180MCU_PR_PATH = @GF180MCU_FD_PR_PATH@
+-GF180MCU_PV_PATH = @GF180MCU_FD_PV_PATH@
 -GF180MCU_IO_PATH = @GF180MCU_FD_IO_PATH@
 -GF180MCU_SC_7T5V0_PATH = @GF180MCU_FD_SC_MCU7T5V0_PATH@
 -GF180MCU_SC_9T5V0_PATH = @GF180MCU_FD_SC_MCU9T5V0_PATH@
 -GF180MCU_SRAM_PATH = @GF180MCU_FD_IP_SRAM_PATH@
```
proppy, line 14: we should also add this line

misc/open_pdks/gf180mcuc/use-sources-from-conda.patch
```diff
@@ -1,18 +1,20 @@
...
 -GF180MCU_OSU_SC_PATH = @GF180MCU_OSU_SC_PATH@
```
proppy, line 15: we should also add this line

proppy:
merging since this seems to be fixing the build.

Issue #315, opened 2023-04-04
refresh patches for failing packages


proppy:
Multiple packages seems to be failing because of outdated patches:
https://github.com/hdl/conda-eda/actions/runs/4602182798/jobs/8130882639

- openroad
- magic
- xls

We should make sure to refresh them to they apply on top of `HEAD`.

/cc @xobs fyi

PR #316, opened 2023-04-06
hls/xls: fix ir_convert rules

proppy:
related to #315


proppy:
reviewed by @hongted offline.

Issue #324, opened 2023-04-27
build failure for most packages


proppy:
It seems that most macosx packages builds fail with the following error:
```
  SpecsConfigurationConflictError: Requested specs conflict with configured specs.
    requested specs: 
      - conda
    pinned specs: 
      - python=3.7
```
https://github.com/hdl/conda-eda/actions/runs/4817053021

and that linux packages builds don't actually build anything:
https://github.com/hdl/conda-eda/actions/runs/4822006760/jobs/8588669967?pr=326

proppy:
Looks like 3.7 support has been dropped:
https://github.com/conda/conda/issues/12436

proppy:
The logs seems to stop after:
```
2023-04-27T16:53:01.0066342Z [32;1m16:53:01[0m | Prepare all required actions
2023-04-27T16:53:01.0255686Z ##[group]Run ./ci
2023-04-27T16:53:01.0256134Z [32;1m16:53:01[0m | env:
2023-04-27T16:53:01.0256487Z [32;1m16:53:01[0m |   ANACONDA_TOKEN: 
2023-04-27T16:53:01.0256846Z [32;1m16:53:01[0m |   ANACONDA_USER: 
2023-04-27T16:53:01.0257223Z [32;1m16:53:01[0m |   DEBIAN_FRONTEND: noninteractive
2023-04-27T16:53:01.0257615Z [32;1m16:53:01[0m |   GHA_MACHINE_TYPE: n2-standard-4
2023-04-27T16:53:01.0258073Z [32;1m16:53:01[0m |   APT_REQUIREMENTS: binutils build-essential csh curl git libgl1 pkg-config sudo wget
2023-04-27T16:53:01.0258517Z [32;1m16:53:01[0m |   PACKAGE: misc/magic
2023-04-27T16:53:01.0258866Z [32;1m16:53:01[0m |   OS_NAME: linux
2023-04-27T16:53:01.0259190Z ##[endgroup]
2023-04-27T16:53:01.1152545Z [32;1m16:53:01[0m | A local action modified in a branch!
2023-04-27T16:53:01.1387668Z [32;1m16:53:01[0m | Post job cleanup.
```
I wonder if this could be related to custom runners?

/cc @PiotrZierhoffer @ajelinski

spnadig:
[...]

Issue #190, opened 2022-03-07
openroad package should be available for more than one version of python


proppy:
currently the CI only build the package for python 3.7 which cause the following error on system that have a more recent version:
```
Specifications:

  - openroad -> python=3.7
  - openroad -> python[version='>=3.7,<3.8.0a0']

Your python: python=3.9
```

We should make sure we also build the package for 3.8, 3.9 and 3.10.

proppy:
Generalizing this bug, this this is also valid for other packages.

PR #326, opened 2023-04-27
ci: do not force python3.7

proppy:
This is a lame attempt at fixing #324.

/cc @xobs @PiotrZierhoffer @ajelinski

### DIFF
--- a/ci/conda-get.sh
+++ b/ci/conda-get.sh
@@ -31,7 +31,6 @@ echo $PATH
 conda info
 conda list
 
-echo "python==3.7.*" > $CONDA_PATH/conda-meta/pinned
 conda update -y conda
 conda install -y anaconda-client
 conda install -y python
--- a/ci/install.sh
+++ b/ci/install.sh
@@ -31,7 +31,7 @@ else
     ADDITIONAL_CHANNELS="litex-hub"
 fi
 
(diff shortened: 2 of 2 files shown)


mithro:
@proppy - We should probably be building against all Conda supported Python versions - https://conda.io/projects/conda/en/latest/user-guide/tasks/manage-python.html

> Conda supports Python 3.8, 3.9, 3.10, and 3.11.

proppy:
> @proppy - We should probably be building against all Conda supported Python versions - 

I think that's orthogonal to the version of python installed on the host; a given package should be able to specify the version of conda to build against in the workflow, see: https://github.com/hdl/conda-eda/blob/master/.github/workflows/Build.yml#L117

proppy:
>  building against all Conda supported Python versions 

Updated #190 to be more generic to track this.

mithro:
Nothing should depend on the host's Python, it should be using the Python from conda?

proppy:
> Nothing should depend on the host's Python, it should be using the Python from conda?

Yes, that's my understanding as well that the conda's python is used after entering the recipe build procress, but I assume it is also installed on the host as some of the bootstrap scripts might require it?

xobs:
I don't think it uses Python from the system, even when doing setup.

A similar-but-orthogonal issue is that a particular Python might not be available on the system. For example, `python3.7` is unavailable on arm64, which is why openroad doesn't build on that platform.

PR #337, opened 2023-06-06
xls,open_pdks,openroad: update patches

proppy:
drop and update patches applied upstream

/cc @QuantamHD


proppy:
Merging as it fixes xls,open_pdks,openroad builds.

Issue #343, opened 2023-07-11
openlane package is broken


proppy:
https://github.com/hdl/conda-eda/actions/runs/5513925973/jobs/10053073416
```
2023-07-10T23:55:37.6853685Z [32;1m23:55:37[0m | [INFO]: Running Synthesis (log: ../_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeh/share/openlane/designs/inverter/runs/RUN_2023.07.10_23.55.33/logs/synthesis/1-synthesis.log)...
2023-07-10T23:55:37.6854481Z [32;1m23:55:37[0m | [ERROR]: during executing yosys script $PREFIX/share/openlane/scripts/yosys/synth.tcl
2023-07-10T23:55:37.6855523Z [32;1m23:55:37[0m | [ERROR]: Log: ../_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeh/share/openlane/designs/inverter/runs/RUN_2023.07.10_23.55.33/logs/synthesis/1-synthesis.log
2023-07-10T23:55:37.6856208Z [32;1m23:55:37[0m | [ERROR]: Last 10 lines:
2023-07-10T23:55:37.6856970Z [32;1m23:55:37[0m | ERROR: TCL interpreter returned an error: can't read "::env(SYNTH_CAP_LOAD)": no such variable
2023-07-10T23:55:37.6857471Z [32;1m23:55:37[0m | child process exited abnormally
2023-07-10T23:55:37.6857844Z [32;1m23:55:37[0m | 
2023-07-10T23:55:37.6858234Z [32;1m23:55:37[0m | [ERROR]: Creating issue reproducible...
```

proppy:
should resolve itself once https://github.com/The-OpenROAD-Project/OpenLane/pull/1892 is merged.

kareefardi:
@proppy does this means that conda builds of OpenLane doesn't use the version of dependencies listed here https://github.com/The-OpenROAD-Project/OpenLane/blob/master/dependencies/tool_metadata.yml ?

proppy:
@kareefardi yes, conda build everything from `HEAD` every might and only publish if the tests (hardening a test design) passes.

proppy:
build failed with the following error:
```
[...]

PR #345, opened 2023-07-20
build packages for python 3.10

proppy:
- github/workflow: add py310 entries
- openroad: use PYTHON_VERSION
- klayout: use pin_compatible
- openlane: update test design
- misc/open_pdks: move python to build dep
- misc/netgen: remove python dep

Related #343 #72

### DIFF
--- a/.github/workflows/Build.yml
+++ b/.github/workflows/Build.yml
@@ -132,8 +132,20 @@ jobs:
       - run: apt -qqy update && apt -qqy install ${{ env.APT_REQUIREMENTS }}
       - uses: ./ci
 
+  yosys-linux-py310:
+    runs-on: [self-hosted, Linux, X64]
+    container: ubuntu:22.04
+    env:
+      PACKAGE: "syn/yosys"
+      OS_NAME: "linux"
+      PYTHON_VERSION: "3.10"
+    steps:
+      - uses: actions/checkout@v3
+      - run: apt -qqy update && apt -qqy install ${{ env.APT_REQUIREMENTS }}
+      - uses: ./ci
+
   yosys-plugins-symbiflow-linux:
-    needs: ["icarus-linux", "yosys-linux-py37", "yosys-linux-py38", "surelog-linux-py37", "surelog-linux-py38"]
+    needs: ["icarus-linux", "yosys-linux-py37", "yosys-linux-py38", "yosys-linux-py310", "surelog-linux-py37", "surelog-linux-py38", "surelog-linux-py310"]
     runs-on: [self-hosted, Linux, X64]
     container: ubuntu:20.04
     env:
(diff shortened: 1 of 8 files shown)


proppy:
@ajelinski @kgugala @umarcor does that sounds like the right thing to do?

pnr/openroad/meta.yaml
```diff
@@ -1,5 +1,7 @@
 # Use `conda-build-prepare` before building for a better version string.
 {% set version = '%s_%04i_%s'|format(GIT_DESCRIBE_TAG|replace('v', '')|replace('-', '')|default('0.X'), GIT_DESCRIBE_NUMBER|int, GIT_DESCRIBE_HASH|default('gUNKNOWN')) %}
+
+{% set python_version = PYTHON_VERSION | default('3.7') %}
```
ajelinski, line 4: Please add `py_suffix` too:
```
{% set py_suffix = 'py%s'|format(python_version|replace('.', '')) %}
```
and use it for build string in lines 25-26:
```
  # string: 20220203_1935_py37
  string: {{ environ.get('DATE_STR') }}_{{ py_suffix }}
```

.github/workflows/Build.yml
```diff
@@ -313,7 +324,19 @@ jobs:
...
+      - uses: ./ci
+
+  symbiyosys-linux-py310:
+    needs: ["yosys-linux-py310"]
+    runs-on: [self-hosted, Linux, X64]
+    container: ubuntu:22.04
+    env:
+      PACKAGE: "formal/symbiyosys"
+      OS_NAME: "linux"
+      PYTHON_VERSION: "3.10"
```
ajelinski, line 339: I see SymbiYosys still can't be built successfully so I'd say let's keep `SKIP` for jobs with Python 3.7/3.8. Unless building is fixed I'm not sure if adding this job for SymbiYosys+Python 3.10 makes sense. It'd be best to fix it but if not maybe let's add the job with `SKIP: "true"`?

proppy: Understood, IIRC build failure used to block publication but it doesn't seem to be the case anymore. Maybe we could revisit removing skipped build in a follow up changes (and let them fails until they are fixed)

ajelinski: Right, it makes sense to just let them fail. Some of them might have been even fixed upstream already so maybe you could drop all `SKIP`s in a separate commit then?

proppy:
@ajelinski does this look good to you?

ajelinski:
> @ajelinski does this look good to you?

The changes strictly related to the Python 3.10 addition LGTM.

Can't say much about other changes but if the packages are built correctly with such requirement changes, `open_pdks.sky130a` needs the new patch and `openlane` should be tested with `-design spm` instead of `-design inverter` then also LGTM.

proppy:
@ajelinski do you prefer that I split those in separate PRs?

The issue is that I couldn't really test those, since they were all related (we can't test python 3.10 without design changes for broken packages, and we can't test the design changes without python 3.10 fixes because of dep conflicts)

ajelinski:
> @ajelinski do you prefer that I split those in separate PRs?
> 
> The issue is that I couldn't really test those, since they were all related (we can't test python 3.10 without design changes for broken packages, and we can't test the design changes without python 3.10 fixes because of dep conflicts)

Sure, it's not the best approach but I guess it's better than the alternatives.

I didn't suggest separate PRs though. I mostly wanted to say that I don't see any problems with the changes but at the same time I just lack knowledge about these packages to say all of them are correct.

Perhaps you could squash the Openroad and `Build.yml` changes into the original commit. Otherwise LGTM. @umarcor ?

proppy:
rebased, and squashed!

Thanks @ajelinski for the quick review.

PR #356, opened 2023-09-28
hls/xls: refresh patches and tests

proppy:
- remove obsolete patches
- fixing tests
- bumping bazel


proppy:
merging as it fixes the build.

Issue #351, opened 2023-08-17
Most package builds are failing


proppy:
https://github.com/hdl/conda-eda/actions/runs/5872663018/job/15924608681

It seems that the job are failing without providing any logging.

/cc @kgugala @ajelinski

proppy:
Seems like the issue resolved itself: https://github.com/hdl/conda-eda/actions/runs/5944971903/job/16123225286

proppy:
Looks like it's happening again.

Issue #198, opened 2022-03-14
Conda-forge and Viper


curtisma:
Have you considered submitting these recipes to Conda-forge rather than having your own channel?   It could handle all the build infrastructure and the recipes would be maintained by the tool owners and/or anyone else interested in helping maintain them.  They also have a well documented process for creating new recipes which could make it easier to support new tools.

I've proposed this approach as part of the [Viper circuit design environment ](http://www.cascode-labs.org/viper/)and [viper-forge](http://www.cascode-labs.org/viper-forge/).

mithro:
Sounds like you just invented the same thing I did --> [bit.ly/edda-conda-eda-spec](https://bit.ly/edda-conda-eda-spec)

mithro:
See also  https://j.mp/edda-status

mithro:
Also some stuff at https://docs.google.com/document/d/11XFnJ0ExBgE1pMQksw0rQerAZo3F83AVIu2YK1pbg1k/edit

curtisma:
haha, cool, want to work together?

curtisma:
I work on more analog / RF IC design but there should be a lot of overlap.

mithro:
BTW Have you seen Google's SkyWater open source PDK (https://github.com/google/skywater-pdk) and the associated no-cost shuttle program for open source designs @ efabless.com?

mithro:
It seems like our work is pretty complimentary too! Our focus has been heavily on the CPU / digital (starting with FPGAs + cross compilers) while your focus has been more on the analog side.

curtisma:
Yeah, I've heard about the Sky water shuttles.  I've been considering putting together and submitting a design.

What's the best way to collaborate?  Do you want to have an online meeting or discuss on one of our GitHub discussion boards perhaps?

proppy:
@curtisma I just happen to have started a PR to add a subset of those package to conda-forge as part of #193, see https://github.com/conda-forge/staged-recipes/pull/18295

curtisma:
@proppy Ok, awesome, do you need help with it?

proppy:
@curtisma rebased the PR and it nows pass the CI for linux: https://github.com/conda-forge/staged-recipes/pull/18295, do you want to take a look before I ping the conda-forge folks about it?

curtisma:
Sure, I can take a look tonight.  That's what I had guessed when I saw it previously

curtisma:
[...]

PR #359, opened 2023-10-18
an attempt to pack openfasoc like openlane with all dependencies

saicharan0112:
Currently OpenFASoC extensively uses conda packages to install all dependencies as part of its installer script and docker images. This is an attempt to pack openfasoc so that it can be installed very easily along with all its dependency packages. But there are still a few more things to do to get the stable package that is built regularly, like below - 
* use conda packages versions info that is stored in a file after the daily check (https://github.com/idea-fasoc/OpenFASOC/blob/main/conda_versions.txt). Need to investigate what changes that need to be done at OpenFASoC side.
* use requirements.txt file to install python libraries that are required for the generators. (https://github.com/idea-fasoc/OpenFASOC/blob/main/requirements.txt)

P.S - I took openlane files under /misc/openlane as reference to add files for OpenFASoC

@msaligane @proppy


saicharan0112:
Please do share your suggestions on how to solve the two problems that I mentioned under points.

proppy:
@saicharan0112 we're seeing issue with the CI https://github.com/hdl/conda-eda/issues/351, we need to make sure to fix those before reviewing this.

saicharan0112:
@proppy Are there any packages that are built with very specific version of its dependent conda packages? I think because of `conda config --set channel_priority strict`, the build is failing to install specific versions of conda packages (see https://github.com/saicharan0112/conda-eda/actions/runs/6846441968/job/18613077303#step:3:1436). 

Also I see that gdsfactory is not in this channel. I know that other pip packages are not available so I have no option but to leave them. However, @joamatab is it ok for you having gdsfactory distributed as a conda package too? If so, I can create one.

proppy:
There is https://github.com/conda-forge/gdsfactory-feedstock, so another alternative would be to move the package to conda-forge (we had such loose plan with #198 and #193)

saicharan0112:
> There is https://github.com/conda-forge/gdsfactory-feedstock, so another alternative would be to move the package to conda-forge (we had such loose plan with #198 and #193), happy to revisit them.

Thanks, where do I need to mention to let the conda-build operation here know that it is present in conda-forge channel? Also is it possible to specify custom versions of conda packages in the metadata of the package? 

About shifting to conda-forge, I am not sure if I can publish the package in conda-forge directly because it required a github release url whereas here I can just mention the github repo url and the branch, and cloning and checking out the branch is well taken here. TBH with conda-eda I see an advantage of getting the tool at the latest available commit (though this is built once in a day, I just need to wait for a single day to get the latest version, if updates are made)

proppy:
> Thanks, where do I need to mention to let the conda-build operation here know that it is present in conda-forge channel? 

It's usually not a good idea to mix both, we would need to either bring other dependencies to conda-forge or bring gdsfactory here based on the existing recipe.

proppy:
> directly because it required a github release url 

maybe openfasoc could do automated release similar to what verible is doing: https://github.com/chipsalliance/verible/releases ?

proppy:
> I just need to wait for a single day to get the latest version

conda-forge also has some automated way to bump version, see what's happening with magic here: https://github.com/conda-forge/magic-feedstock/pull/101

PR #361, opened 2023-10-23
pnr/openroad: fix patch

proppy:
refreshed the boost patch again: https://github.com/proppy/conda-eda/blob/437ca9a3fc534c82f822f3e552857453086215c0/pnr/openroad/remove-boost-span-deps.patch

@QuantamHD do you think it would be worth landing something cleaner upstream? e.g: make the dependency on boost-span optional (since it's only used in tests)


proppy:
Merging this this fixed the ci build: https://github.com/hdl/conda-eda/actions/runs/6621655996/job/17985949610?pr=361

PR #362, opened 2023-10-24
hls/xls: add macosx build

proppy:
/cc @leary @xobs


proppy:
closing this, as osx/intel not a platform supported by the xls project.

PR #363, opened 2023-10-24
misc/openlane: add verilator dep

### LINKED ISSUE #352: openlane package is broken
https://github.com/hdl/conda-eda/actions/runs/5944971903/job/16125504708

proppy:
should fix #352


proppy:
Merging as it does fix the build.

PR #364, opened 2023-11-02
add klayout 3.11 and 3.12

joamatab:


proppy:
merging since CI seems to pass👍

Thanks!

Issue #365, opened 2024-01-22
OpenROAD version string missing `git describe` output


ubfx:
It seems like the version string compiled into OpenROAD is missing the `git describe` output:

```bash
$ openroad -v
OpenROAD e89829335596b351ce665dcc3b73619c0b191c14 
```

while it should look something like `OpenROAD v2.0_... `. Interestingly, the anaconda package version string contains the git describe output: `openroad 2.0_3078_g4e7b85576`.

cc @proppy

PR #374, opened 2024-02-23
OpenROAD: Fix version string in binary

ubfx:
Currently  `git describe` is failing within the OpenROAD build which leads to the version number compiled into the binary falling back to a hash which doesn't work well for version checking.

https://github.com/hdl/conda-eda/actions/runs/8012258777/job/21887187187#step:5:2091
```
  23:23:44 | CMake Warning at CMakeLists.txt:98 (message):
  23:23:44 |   OpenROAD git describe failed, using sha1 instead
```

The reson seems to be that cmake is called with `-DCMAKE_FIND_ROOT_PATH` such that `find_package(Git)` fails. Easiest solution seems to be to just add git to the build requirements, which this patch does. Alternatively we could probably pass `GIT_EXECUTABLE` to cmake.

Fix https://github.com/hdl/conda-eda/issues/365

### DIFF
--- a/pnr/openroad/meta.yaml
+++ b/pnr/openroad/meta.yaml
@@ -39,13 +39,14 @@ requirements:
     - cmake >=3.19
     - swig 4.0.2
     - bison
+    - git >= 2.0          # cmake needs to be able to find git to establish a version string
     - {{ compiler('c') }}
     - {{ compiler('cxx') }}
     - {{ cdt('libx11-devel') }}         # [linux]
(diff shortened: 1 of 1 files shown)


ubfx:
Looks like this worked 🎉 

https://github.com/hdl/conda-eda/actions/runs/8015700461/job/21896297577?pr=374#step:5:2239
```
06:57:23 | -- OpenROAD version: 2.0-12327-g933ae804c
```

image

cc @proppy

proppy:
Thanks for figuring that out!

PR #375, opened 2024-02-23
OpenROAD: Do not ignore run_export from fmt

### LINKED ISSUE #360: Openroad missing library issue
Running a conda install for openroad==2.0_10145_g0de6f5131 on ubuntu 20.04 seems to fail due to the absence of libfmt v8 on the distro
The following error is thrown when it is used

`miniconda3/bin/openroad: error while loading shared libraries: libfmt.so.8: cannot open shared object file: No such file or directory`
Is there any workaround for Ubuntu 20.04?

ubfx:
It seems like on Ubuntu 20.04 (reported here https://github.com/hdl/conda-eda/issues/360) and CentOS 7 (tested by me), we have to manually `conda install fmt` for the OpenROAD binary from our package to work.

I'm not sure this will fix it but maybe we should try removing `fmt` from `ignore_run_exports_from`?

Fix https://github.com/hdl/conda-eda/issues/360 (hopefully)

### DIFF
--- a/pnr/openroad/meta.yaml
+++ b/pnr/openroad/meta.yaml
@@ -31,7 +31,6 @@ build:
     - CI
   ignore_run_exports_from:
     # header-only libraries
-    - fmt
 
 requirements:
   build:


ubfx:
cc @proppy, if you have time: I think this and https://github.com/hdl/conda-eda/pull/373 could be merged and then the package should be good for now :)

PR #378, opened 2024-06-18
sv2v: remove stack version upper bound

zachjs:
`stack` version 2.5 doesn't support the default version of GHC now used to build sv2v. This change attempts to fix the build of sv2v by removing the included `stack` version upper bound.


zachjs:
@mithro When you get the chance, could you please approve the workflow run here?

